PR #123, opened 2020-04-01
Stores last block on db on fresh bootstrap for Watcher and Responder

### LINKED ISSUE #117: Bootstrapping with no "last_block_watcher" stops teosd from starting
Encountered a small bug, I think. I had run the watchtower daemon before & added a test appointment, so the second time it started bootstrapping from backed up data in teosd.py. But since I was using regtest, no blocks had been mined and no "last_block" had been added to the database.         

So the program trying to retrieve the last block on line 115 returns "None." Such a block doesn't exist, so the server shuts down

sr-gi:
The last known block was not stored in the db for the `Watcher` and `Responder` on fresh bootstrap until a block was received. This caused problems if stopping the tower before hearing about a block, specially in testing.

Fixes #117

- Stores `last_known_block` in the db for both the `Watcher` and `Responder` on fresh bootstrap (before `do_watch` main loop)
- Adds `last_known_block` to `Watcher` to deal with forks in the future.

### DIFF
--- a/teos/watcher.py
+++ b/teos/watcher.py
@@ -54,6 +54,7 @@ class Watcher:
         signing_key (:mod:`PrivateKey`): a private key used to sign accepted appointments.
         max_appointments (:obj:`int`): the maximum amount of appointments accepted by the ``Watcher`` at the same time.
         expiry_delta (:obj:`int`): the additional time the ``Watcher`` will keep an expired appointment around.
+        last_known_block (:obj:`str`): the last block known by the ``Watcher``.
 
     Raises:
         ValueError: if `teos_sk_file` is not found.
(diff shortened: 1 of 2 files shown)


teos/watcher.py
```diff
@@ -70,6 +71,7 @@ def __init__(self, db_manager, block_processor, responder, sk_der, max_appointme
         self.max_appointments = max_appointments
         self.expiry_delta = expiry_delta
         self.signing_key = Cryptographer.load_private_key_der(sk_der)
+        self.last_known_block = db_manager.load_last_block_hash_responder()
```
orbitalturtle, line 74: Should this be db_manager.load_last_block_hash_watcher()?

sr-gi: Indeed, bad copy paste :(

sr-gi:
Can't reproduce with `regtest` at height 0.

Is it exactly the same error? Can you add the steps to reproduce?

PR #124, opened 2020-04-02
Adds user authentication

sr-gi:
Adds user-side authentication to the tower. Users now have to register in order to further interact with `teos`.

Registration is free for now, so users simply need to send a registration request to the register endpoint (teos_cli.py register or send a json POST to `/register` containing a public_key field with a compressed public key).

A new entity has been created to deal with the user access to the tower: the `Gatekeeper`. The Gatekeeper registers the users and checks that every request comes from a registered user, if not, the request are dropped. User authentication is based on EC recovery (#103). The `DBManager` has been split into a parent class (`DBManager`) and two child classes: `AppointmentDBM` and `UsersDBM`. The parent has kept the shared functionality, whereas the children have functions related to their purpose (keep data of the `appointments` or `users`). The `Gatekeeper` is in charge os the `UserDBM`.
On register a number of `appointment_ slots` will be given to the user, and every appointment will consume slots depending on its size. Updates won't consume additional slots as long as they are within the same bounds of the old appointment. Updates of smaller size may free slots whereas updates of bigger size may fill additional slots.

The `API` has been updated to add the `register` endpoint and modify the `get_appointment` endpoint to work via `POST` instead of `GET`. Furthermore, the `add_appointment` endpoint has been changed from `\` to `\add_appointment` for consistency.

Appointment from the same user sharing the same locator are now seen as updates, and will overwrite the old appointments. Therefore, a single appointment is kept per user and locator. 
In order to minimise the data kept in the database regarding users, the uuid computation has been changed. `uuids` are not random anymore, but based on both the `locator` and the `user_pk`.  The same `uuid` is shared by updates from the same user, and `get_appointment` requests allow the tower to recompute the `uuid`, so no uuid:user_pks maps have to be stored (at least not for now).

The `Inspector` has been simplifier to return custom exception on inspect failures, making it's design way cleaner (we should refactor more code following this approach).
Further restrictions based on appointment size have been lifted.

Finally, docs and comments have been revised and updated.


CONTRIBUTING.md
```diff
@@ -8,6 +8,15 @@ We use [black](https://github.com/psf/black) as our base code formatter with a l
...
 black --line-length=120 {source_file_or_directory}
 ```
+
+In additon, we use [flake8](https://flake8.pycqa.org/en/latest/) to detect style issues with the code:
+
+```bash
+flake8 --max-line-length=120 {source_file_or_directory}
+```
+
+ Not all outputs from flake8 are mandatory. For instance, splitting **bullet points in docstrings (E501)** will cause issues when generating the docuementation, so we will leave that longer than the line lenght limit . Another example are **whitespaces before colons in inline fors (E203)**. `black` places them in that way, so we'll leave them like that.
```
bigspider, line 18: typos: documentation, length

common/cryptographer.py
```diff
@@ -1,19 +1,19 @@
...
-from common.tools import check_sha256_hex_format
+from common.tools import is_256b_hex_str
 
 LN_MESSAGE_PREFIX = b"Lightning Signed Message:"
 
 
 def sha256d(message):
     """
-    Compute the sha245d (double sha256) of a given by message.
+    Compute the sha256 (double sha256) of a given by message.
```
bigspider, line 16: Should just say `Computes the double sha256 of a given message`, I think

teos/appointments_dbm.py
```diff
@@ -0,0 +1,508 @@
...
+        - ``WATCHER_PREFIX``, defined as ``b'w``, is used to store :obj:`Watcher <teos.watcher.Watcher>` appointments.
+        - ``RESPONDER_PREFIX``, defines as ``b'r``, is used to store :obj:`Responder <teos.responder.Responder>` trackers.
+        - ``WATCHER_LAST_BLOCK_KEY``, defined as ``b'bw``, is used to store the last block hash known by the :obj:`Watcher <teos.watcher.Watcher>`.
+        - ``RESPONDER_LAST_BLOCK_KEY``, defined as ``b'br``, is used to store the last block hash known by the :obj:`Responder <teos.responder.Responder>`.
+        - ``LOCATOR_MAP_PREFIX``, defined as ``b'm``, is used to store the ``locator:uuid`` maps.
+        - ``TRIGGERED_APPOINTMENTS_PREFIX``, defined as ``b'ta``, is used to stored triggered appointments (appointments that have been handed to the :obj:`Responder <teos.responder.Responder>`.)
+
+    Args:
+        db_path (:obj:`str`): the path (relative or absolute) to the system folder containing the database. A fresh
+            database will be create if the specified path does not contain one.
```
bigspider, line 36: typo "will be created"

teos/appointments_dbm.py
```diff
@@ -0,0 +1,508 @@
...
+    def load_watcher_appointments(self, include_triggered=False):
+        """
+        Loads all the appointments from the database (all entries with the ``WATCHER_PREFIX`` prefix).
+
+        Args:
+            include_triggered (:obj:`bool`): whether to include the appointments flagged as triggered or not. ``False``
+                by default.
+
+        Returns:
+            :obj:`dict`: A dictionary with all the appointments stored in the database. An empty dictionary is there
```
bigspider, line 151: typo "if there are none"; or you could remove the last sentence - I find it redundant.

teos/appointments_dbm.py
```diff
@@ -0,0 +1,508 @@
...
+
+        self.db.put((TRIGGERED_APPOINTMENTS_PREFIX + uuid).encode("utf-8"), "".encode("utf-8"))
+        logger.info("Flagging appointment as triggered", uuid=uuid)
+
+    def batch_create_triggered_appointment_flag(self, uuids):
+        """
+        Creates a flag that signals that an appointment has been triggered for every appointment in the given list
+
+        Args:
+            uuids (:obj:`list`): a list of identifier for the appointments to flag.
```
bigspider, line 456: typo "identifiers"

teos/builder.py
```diff
@@ -94,8 +98,9 @@ def populate_block_queue(block_queue, missed_blocks):
     @staticmethod
     def update_states(watcher, missed_blocks_watcher, missed_blocks_responder):
         """
-        Updates the states of both the :mod:`Watcher <teos.watcher.Watcher>` and the :mod:`Responder <teos.responder.Responder>`.
-        If both have pending blocks to process they need to be updates at the same time, block by block.
+        Updates the states of both the :mod:`Watcher <teos.watcher.Watcher>` and the
+        :mod:`Responder <teos.responder.Responder>`. If both have pending blocks to process they need to be updates at
```
bigspider, line 102: typo "updated"

teos/gatekeeper.py
```diff
@@ -0,0 +1,117 @@
+from common.tools import is_compressed_pk
+from common.cryptographer import Cryptographer
+
+
+class NotEnoughSlots(ValueError):
+    """Raise this when trying to subtract more slots than a user has available"""
```
bigspider, line 6: I'd write the descriptions as "Raised when...", so it's equally good for the programmer raising it and for the one that might want to catch it.

teos/gatekeeper.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Args:
+            user_pk(:obj:`str`): the public key that identifies the user (33-bytes hex str).
+
+        Returns:
+            :obj:`int`: the number of available slots in the user subscription.
+        """
+
+        if not is_compressed_pk(user_pk):
+            raise ValueError("provided public key does not match expected format (33-byte hex string)")
```
bigspider, line 48: capital "P"

teos/users_dbm.py
```diff
@@ -0,0 +1,128 @@
...
+    def load_user(self, user_pk):
+        """
+        Loads a user record from the database using the ``user_pk`` as identifier.
+
+        Args:
+
+            user_pk (:obj:`str`): a 33-byte hex-encoded string identifying the user.
+
+        Returns:
+            :obj:`dict`: A dictionary containing the appointment data if they ``key`` is found.
```
bigspider, line 78: Typo "they"

test/teos/e2e/test_basic_e2e.py
```diff
@@ -24,51 +24,89 @@
...
+
+
+def add_appointment(appointment_data, sk=cli_sk):
+    return teos_cli.add_appointment(
+        appointment_data, sk, teos_pk, teos_base_endpoint, cli_config.get("APPOINTMENTS_FOLDER_NAME")
+    )
+
+
+def test_commands_non_registered(bitcoin_cli, create_txs):
+    # All commands should failed if the user is not registered
```
bigspider, line 57: typo "should failed"

test/teos/unit/test_api.py
```diff
@@ -1,202 +1,521 @@
...
-    appointment_locators = [appointment["locator"] for appointment in received_appointments]
 
-    assert new_appt_data["appointment"]["locator"] in appointment_locators and len(received_appointments) == 1
-    assert all([status == "dispute_responded" for status in appointment_status]) and len(appointment_status) == 1
+def test_request_appointment_not_registered_user(client):
+    # Not registered users have no associated appointments, so this should fail
+    tmp_sk, tmp_pk = generate_keypair()
+
+    # The tower is designed so a not found appointment and a request from a non-registered user return the same error to
+    # prevent proving.
```
bigspider, line 391: "probing", I think

PR #132, opened 2020-04-07
Adds API HOST and PORT as configurable parameters

### LINKED ISSUE #131: teos port bind not settable from cli
daemon bind interface is not settable which complicates running teos remotely (or in docker) as it only listens on loopback. It should be a configurable part of the setup - both with cli arguments and .conf file

sr-gi:
Renames them to API_CONNECT and API_PORT for consistency.

Close #131

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -15,8 +15,8 @@ Refer to [INSTALL.md](INSTALL.md)
 	
 #### Global options
 
-- `-s, --server`:	API server where to send the requests. Defaults to 'localhost' (modifiable in conf file).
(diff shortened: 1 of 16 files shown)


sr-gi:
Can you try if it's easier now @aljazceru?

Feel free to comment on the renaming too, I'm normally pretty bad at this stuff :)

aljazceru:
@sr-gi  yes, this is great now, merged to my fork and tested it 

maybe API_BIND would be better as this controls to what interface/address teos binds not where it connects

PR #139, opened 2020-05-05
Watchtower plugin

sr-gi:
Adds a c-lightning plugin based on the commitment_revocation hook from https://github.com/ElementsProject/lightning/pull/3601.

The plugin has the following methods:

- `registertower <tower_id>` : registers the user id (compressed public key) with a given tower.
- `getappointment <tower_id><locator>`: queries an appointment to a given tower.
- `list_towers`: lists all the registered towers.
- `gettowerinfo <tower_id>`: gets all the locally stored data about a given tower.
- `retrytower <tower_id>`: tries to send pending appointment to a (previously) unreachable tower.

The plugin also has an implicit `add_appointment` method triggered via the  `commitment_revocation` hook.

The current approach has free registration and sends the data to all registered towers. There is no management of the tower subscriptions at the moment, so if the plugin runs out of available slots for a given tower the user will need to request more (via `registertower`).

Data sent to towers that are unreachable / there's a subscription error are backed up in `pending_appointments`. A retry mechanism is fired by default with an exponential backoff strategy. If the retry fails, it can be triggered manually via `retrytower`.

Appointments flagged as invalid by the tower are stored in `invalid_appointments`. This should never happen to honest clients.

If the tower is caught misbehaving, for instance sending an invalid signature, it will be flagged as misbehaving and the proof will be stored in `misbehaving_proof`.

All data is stored on disk in a LevelDB.

### Pending

- More tests may be added before closing
- Test with multiple towers at the same time


sr-gi:
Ping @cdecker @darosior some tests are still missing, but I think the plugin is on a good shape now.  Feel free to give it a try.

I can setup a live instance of the tower so it's easier, just let me know.

PR #148, opened 2020-05-20
System monitor

orbitalturtle:
This PR puts together a first draft of a system monitor for a Teos watchtower. It can be run once Elasticsearch and Kibana are running to produce a Kibana dashboard with several graphs displaying Teos data.

Would love some feedback if anyone has time!

I also have my eyes on the following improvements:
- Right now each time the program boots up, in order to gather the most recent log data, it wipes out the data that was entered into Elasticsearch entirely and refills it with the data from the log file. Would probably be better to track how many logs have been entered so far and when the program starts up, start where it left off.
- The data could be updated in real-time (maybe Logstash can help here? I'm not super familiar)
- I might be missing something, but right now it seems the regtest Teos log data and mainnet log data are mixed in to the same log file, which would muddle the data in the dashboard. So perhaps would be good to make separate files for each.
- Testing and exception handling are very minimal so far 
- There's probably much more that can be graphed :-) esp. time series data


sr-gi:
Supper excited about this 🚀 I'll give it a look asap.

yahgwai:
Looks cool, always nice to see some dashboards :)

I think you might be able to do without your own custom loading code though. You can take two basic approaches to loading logs: handler or agent.

The handler approach is to add an additional logging stream. You already have a console and file stream:
```
self.f_logger.info(self._create_file_message(msg, **kwargs))
self.c_logger.info(self._create_console_message(msg, **kwargs))
```
so you would just add a third stream for elasticsearch. To do this you'd want to use a 3rd party lib as it would do a better job of keeping track of location (in case of crash), and avoiding blocking io. I dont work with python so I dont know how to assess the quality of the lib, but this seems ok: https://github.com/cmanaha/python-elasticsearch-logger

The agent approach is the one you've taken already, but you can use an existing agent instead of writing your own. Filebeat (https://www.elastic.co/beats/filebeat) will keep track of files and locations, and you can use logstash if you need transforms (https://www.elastic.co/logstash).

There's good discussion comparing the two approaches here (https://github.com/cmanaha/python-elasticsearch-logger/issues/44/) with the author `python-elasticsearch-logger`

yahgwai:
You have potential problem with overlapping fields of different types. Elasticsearch wont accept the same field with a type different to it's mapping. So if you have a two logs like:
```
logger.info(message, testy="testy")
logger.info(message, testy=10)
```
the latter would be rejected. 

I havent checked to see if there are any instances of this issue, just flagging it as potential problem.

sr-gi:
Starting to check this finally, sorry for taking so long :(

monitor/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+
+MONITOR_DEFAULT_CONF = {
+    "ES_HOST": {"value": "", "type": str},
+    "ES_PORT": {"value": 9200, "type": int},
+    "KIBANA_HOST": {"value": "localhost", "type": str},
+    "KIBANA_PORT": {"value": "9243", "type": int},
+    "CLOUD_ID": {"value": "", "type": str}, 
+    "AUTH_USER": {"value": "user", "type": str},
+    "AUTH_PW": {"value": "password", "type": str},
+}
```
sr-gi, line 14: Some defaults are missing here. Feel free to leave some empty if you feel like it (like remote auths), but one should at least be set so the code can run with defaults:

KIBANA_PORT should be `int` in defaults
CLOUD_ID has no default
ES_HOST has no default

Also, the default port pot Kibana is 5601 for me, isn't it for you?

monitor/data_loader.py
```diff
@@ -0,0 +1,279 @@
...
+         self.es_host = es_host
+         self.es_port = es_port
+         self.es_cloud_id = cloud_id
+         self.es_auth_user = auth_user
+         self.es_auth_pw = auth_pw
+         if self.es_host is not "":
+             self.es = Elasticsearch([
+                 {'host': self.es_host, 'port': self.es_port}    
+             ])     
+         elif cloud_id is not "":
```
sr-gi, line 54: This is missing a default else (in case no host not cloud_id is provided). This won't be necessary as long as the config is properly checked before feeding info to the function so it fails early.

monitor/data_loader.py
```diff
@@ -0,0 +1,279 @@
+import json
+import os
```
sr-gi, line 2: Never used

monitor/data_loader.py
```diff
@@ -0,0 +1,279 @@
...
+import time
+
+from elasticsearch import Elasticsearch, helpers
+from elasticsearch.client import IndicesClient
+from elasticsearch.helpers.errors import BulkIndexError
```
sr-gi, line 7: Never used either

monitor/monitor_start.py
```diff
@@ -0,0 +1,49 @@
...
+logger = Logger(actor="System Monitor Main", log_name_prefix=LOG_PREFIX)
+
+def main(command_line_conf):
+    logger.info("Setting up the system monitor.")
+
+    # Pull in Teos's config file to retrieve some of the data we need.
+    conf_loader = ConfigLoader(DATA_DIR, CONF_FILE_NAME, DEFAULT_CONF, command_line_conf)
+    conf = conf_loader.build_config()
+
+    max_users = conf.get("DEFAULT_SLOTS")
```
sr-gi, line 20: Not completely sure what max_users is used for, but you're using `DEFAULT_SLOTS`, which is used to define the default number of slots a users gets for a free subscription.

monitor/monitor_start.py
```diff
@@ -0,0 +1,49 @@
...
+    dataLoader.start()
+
+    kibana_host = mon_conf.get("KIBANA_HOST")
+    kibana_port = mon_conf.get("KIBANA_PORT")
+
+    visualizer = Visualizer(kibana_host, kibana_port, auth_user, auth_pw, max_users)
+    visualizer.create_dashboard()
+
+if __name__ == "__main__":
+    # TODO:
```
sr-gi, line 45: I guess this can be removed

monitor/sample-monitor.conf
```diff
@@ -0,0 +1,14 @@
...
+LOG_FILE = teos.log 
+CONF_FILE_NAME = teos.conf
+
+[Elastic]
+ES_HOST = localhost 
+ES_PORT = 9200
+KIBANA_HOST = localhost
+KIBANA_PORT = 9243
+
+# CLOUD_ID = System_monitor:sdiohafdlkjfl
```
sr-gi, line 12: Can this last ones be deleted?

I guess there should be a kind of priority or check in the code.

If we go with priority, just prioritise one kind of connection over the other. If there's a check, check that only one is set, otherwise exit with an error.

PS: In case you decide to go with priority, at least warning the user so they are aware of both settings being active.

monitor/visualizer.py
```diff
@@ -0,0 +1,95 @@
+import json
+import requests
+
+from monitor.searcher import LOG_PREFIX
```
sr-gi, line 4: monitor.searcher does not exists. Watch out if you try to import it from monitor_start since that'll be a circular dependency and the code will fail (the error message in this cases is pretty vague and tricky t figure out xD)

monitor/data_loader.py
```diff
@@ -0,0 +1,279 @@
...
+             )
+         self.index_client = IndicesClient(self.es)
+         self.log_path = log_file 
+         self.api_host = api_host
+         self.api_port = api_port
+
+    def start(self):
+        """Loads data to be visualized in Kibana"""
+
+        self.delete_index("logs")
```
sr-gi, line 67: This fails on a fresh run. I guess it is trying to delete old data. Can you check that it exists first?

orbitalturtle:
Awesome thanks for the feedback @yahgwai and @sr-gi, I'll look into this over the weekend

monitor/monitor_start.py
orbitalturtle: Ohhhhhhh I see. Lol, I'm not sure how I reached this conclusion, but I was thinking that was the maximum number of users a watchtower supported. But I see, there's a maximum # of appointments, but not users. Will rework the graph I made with that data bit based on that

monitor/data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+        Deletes the chosen index of Elasticsearch.
+    
+        Args:
+            index (:obj:`str`): The ES index to delete.
+        """
+
+        results = self.index_client.delete(index)
+
+    # For testing purposes...
+    def search_logs(self, field, keyword, index):
```
sr-gi, line 227: This does not seem to be used anymore. If the only purpose is testing, maybe it could be part of a local temp file or the testing suite.

monitor/data_loader.py
```diff
@@ -0,0 +1,264 @@
...
+    
+        body = {
+            "query": {"match": {"doc.{}".format(field): keyword}} 
+        }
+        results = self.es.search(body, index) 
+    
+        return results
+    
+    
+    def get_all_logs(self):
```
sr-gi, line 248: This method does not seem to be used anywhere.

monitor/visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+
+logger = Logger(actor="Visualizer", log_name_prefix=LOG_PREFIX)
+
+
+class Visualizer:
+    def __init__(self, kibana_host, kibana_port, max_users):
+        self.kibana_endpoint = "http://{}:{}".format(kibana_host, kibana_port)
+        self.saved_obj_endpoint = "{}/api/saved_objects/".format(self.kibana_endpoint)
+        self.space_endpoint = "{}/api/spaces/space/".format(self.kibana_endpoint)
+        self.headers = headers = {
```
sr-gi, line 16: There's a double variable assignment there

monitor/visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+        self.saved_obj_endpoint = "{}/api/saved_objects/".format(self.kibana_endpoint)
+        self.space_endpoint = "{}/api/spaces/space/".format(self.kibana_endpoint)
+        self.headers = headers = { 
+            "Content-Type": "application/json",
+            "kbn-xsrf": "true"
+        }
+        self.max_users = max_users
+
+    def create_dashboard(self):
+        index_pattern = None
```
sr-gi, line 23: This initialisations do not seem necessary. A dict returns `None` if a value is not found when calling get.

monitor/visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+        image_url = None
+
+        with open(os.getcwd() + '/monitor/kibana_data.json') as json_file:
+            data = json.load(json_file)
+            index_pattern = data.get("index_pattern")
+            visualizations = data.get("visualizations")
+            dashboard = data.get("dashboard") 
+            image_url = data.get("imageUrl") 
+        
+        index_id = None
```
sr-gi, line 35: Same here, it does not need init.

PR #152, opened 2020-06-05
Docker container for teos

aljazceru:
Adds docker building for `teos`.

Instructions of how to run it can be found on the `README`.

The RPC and ZMQ connections from `teos` to the local node works, as well as sending command from the host to the container via HTTP.

The only current downside is that `get_all_appointments` cannot be used from the host in Windows and Mac, since it expects requests to come from localhost atm. I'll open a ticket to find an alternative solution to that.


sr-gi:
I'll take it from here!

sr-gi:
Some testing would be appreciated if you have time (@aljazceru @yahgwai @bigspider @orbitalturtle):

- Build from the Docker file
- Create a block in regtest and check that the towers gets it
- Try to send a request to the tower from outside the container (this one is currently not working for me)

Notice that the system build on mainnet by default, so you'll need to pass the proper ENV variables to change that. Feel free to ping me about it if you hit a wall.

sr-gi:
It'll be also nice if you can also give it a try @stonecoldpat, since you have a Mac. That would help me figure out if the connection issue is local or not.

sr-gi:
Fixed the issue w/ Docker on Mac regarding sending requests to the container.

bigspider:
Worked for me :+1:

sr-gi:
> Worked for me 👍
> 
> Only drawback compared to the instructions is that I had to not only specify that the network is "regtest", but also the port, despite it being the default one; one would expect that it defaults to the default port for that network, perhaps.

Should be fixed by the last set of commits.

RiccardoMasutti:
@nicolasdorier new btcpayserver additional fragment? :)

PR #154, opened 2020-06-09
teos+cli: adds functionality to delete an appointment

sr-gi:
Delete appointment WIP. Will add more info once it moves from draft.

PR #157, opened 2020-06-15
Delete appointment request

sr-gi:
Continuing #154  here


sr-gi:
Leaving this on hold until a good solution for #158 is found

PR #163, opened 2020-06-29
Updates the serialisation format for appointment signatures

sr-gi:
The current serialisation format for appointment signatures differ from the one in BOLT13 (`to_self_delay` and `encrypted_blob` are swapped). Moreover, for appointment accepted messages, the tower does not need to sign the same fields as the user plus the user signature, since signing the latter will implicitly include the former.

This PR modifies both serialisation formats as follows:

Appointment (signed by the user):

`locator | encrypted_blob | to_self_delay`

Receipt (signed by the tower) :

`user_signature | start_block`

Main changes:

`cli`:
- Splits the `add_appointment` method in two: `create_appointment` and `add_appointment`. Also adds stricter checks to the user provided data.

`cli` and `watchtower-plugin`:
- Updates the way the client side computes the data to verify the tower signature (`appointment.serialize` -> `appointment.create_receipt`)

`common`:
- `create_receipt` has been added to `Appointment` as it is used as data to be signed by the tower (instead of `appointment.serialize` as the user does).

`teos`:
- Goes back to using `Appointment` instead of `ExtendedAppointment` in the `api` and `inspector`. The `ExtendedAppointment` is constructed in the `watcher`, since he is the one having all the required data to do so, and also the one that needs to deal with it.
- Adds `start_block` and `user_signature` to the serialization method of `ExtendedAppointment` so that data is also saved on the appointments database.
- Updates the data the tower signs on appointment acceptance (`appointment.serialize` -> `appointment.create_receipt`).

`tests`:

- Updates unit and e2e tests accordingly
- Remove old (unused) tests variables and dictionary fields
- Updates `getblockcount` from `bitcoind_mock` since it had an off-by-one error on computing the height.

### DIFF
--- a/teos/extended_appointment.py
+++ b/teos/extended_appointment.py
@@ -5,8 +5,8 @@ class ExtendedAppointment(Appointment):
     """
     The :class:`ExtendedAppointment` contains extended information about an appointment between a user and the tower.
 
-    While it shares some information with an :class:`Appointment <common.appointment.Appointment>`, it also
-    includes information relevant to the tower, such as the ``user_id``, ``user_signature`` and ``start_block``.
+    It extends the :class:`Appointment <common.appointment.Appointment>` with information relevant to the tower, such
+    as the ``user_id``, ``user_signature`` and ``start_block``.
 
     Args:
         locator (:obj:`str`): A 16-byte hex-encoded value used by the tower to detect channel breaches. It serves as a
--- a/teos/extended_appointment.py
+++ b/teos/extended_appointment.py
@@ -2,13 +2,33 @@ from common.appointment import Appointment
 
 
 class ExtendedAppointment(Appointment):
-    def __init__(self, locator, to_self_delay, encrypted_blob, user_id):
-        super().__init__(locator, to_self_delay, encrypted_blob)
+    """
+    The :class:`ExtendedAppointment` contains extended information about an appointment between a user and the tower.
+
+    While it shares some information with an :class:`Appointment <common.appointment.Appointment>`, it also
+    includes information relevant to the tower, such as the ``user_id``, ``user_signature`` and ``start_block``.
+
+    Args:
+        locator (:obj:`str`): A 16-byte hex-encoded value used by the tower to detect channel breaches. It serves as a
+            trigger for the tower to decrypt and broadcast the penalty transaction.
+        encrypted_blob (:obj:`str`): An encrypted blob of data containing a penalty transaction. The tower will decrypt
+            it and broadcast the penalty transaction upon seeing a breach on the blockchain.
+        to_self_delay (:obj:`int`): The ``to_self_delay`` encoded in the ``csv`` of the ``to_remote`` output of the
+            commitment transaction that this appointment is covering.
+        user_id (:obj:`str`): the public key that identifies the user (33-bytes hex str).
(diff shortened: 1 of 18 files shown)


teos/extended_appointment.py
```diff
@@ -2,13 +2,33 @@
 
 
 class ExtendedAppointment(Appointment):
-    def __init__(self, locator, to_self_delay, encrypted_blob, user_id):
-        super().__init__(locator, to_self_delay, encrypted_blob)
+    """
+    The :class:`ExtendedAppointment` contains extended information about an appointment between a user and the tower.
+
+    While it shares some information with an :class:`Appointment <common.appointment.Appointment>`, it also
```
bigspider, line 8: "it shares some information" seems to suggests that neither class is subclass of the other. I'd say instead "It extends the Appointment with information relevant to the tower, such as..."

teos/extended_appointment.py
```diff
@@ -36,11 +56,20 @@ def from_dict(cls, appointment_data):
 
         appointment = Appointment.from_dict(appointment_data)
         user_id = appointment_data.get("user_id")
+        user_signature = appointment_data.get("user_signature")
+        start_block = appointment_data.get("start_block")
 
-        if not user_id:
-            raise ValueError("Wrong appointment data, user_id is missing")
+        if any(v is None for v in [user_id, user_signature, start_block]):
```
bigspider, line 62: Could be more compact with:
```python
if None in [user_id, user_signature, start_block]:
```

Although here perhaps it would be even better to check each field individually to give more precise error messages?

test/cli/unit/test_teos_cli.py
```diff
@@ -144,7 +179,7 @@ def test_load_keys():
     with open(empty_file_path, "wb"):
         pass
 
-    # Now we can test the function passing the using this files
+    # Now we can test the function using this files
```
bigspider, line 182: typo, "these files"

test/teos/unit/test_inspector.py
```diff
@@ -185,7 +183,7 @@ def test_inspect(run_bitcoind):
     appointment = inspector.inspect(appointment_data)
 
     assert (
-        type(appointment) == ExtendedAppointment
```
bigspider, line 188: I think `isinsance(appointment, Appointment)` is considered more idiomatic?

teos/extended_appointment.py
sr-gi: I was thinking there was a good reason for this, but turns out there is none. This comes from a refactor of `all(v is not None for v in ...` which I actually overlooked at the time ([guess who suggested it](https://github.com/talaia-labs/python-teos/pull/57#discussion_r343459366) :P).

I'll change this and the equivalent in `Appointment` in a separate PR.

test/teos/unit/test_inspector.py
sr-gi: Agreed. However, looking at this closely, it can be removed. That check comes from an older version of `Inspector.inspect` where the method had multi-return (Appointment or None), currently the method raises an exception if something goes wrong.

teos/extended_appointment.py
bigspider: Lol. Well, my reviews improve jointly with my python :P

Issue #169, opened 2020-07-07
Have separate data folders for mainnet, testnet and regtest


bigspider:
Currently, the data folder for logs and appointments is the same (e.g. `~/.teos`) regardless of which network teos is running against. That creates a risk of running teos with the data folder of a different network than the intended one.

Mimicking bitcoind, it would be better to have different data folders for different network, by adding a suffix for non-mainnet networks. For example:

- ~/.teos contains teos.conf and is the data folder for mainnet
- ~/.teos/regtest is the data folder for regtest
- ~/.teos/testnet is the data folder for testnet
- similarly for signet, I guess

sr-gi:
Don't bother about signet for now since there is no coverage atm.

Issue #170, opened 2020-07-07
Tower key management


sr-gi:
Currently, the tower keypair (identity + sk) have to be manually created on first tower boostrap. However, this could easily be automated so the user does not have to bother with it (better UX).

Furthermore, options to overwrite the key pair and to import and external one (either to create it independently or to reuse it from an old build) should be added.

sr-gi:
I think we can squeeze this in  [v0.0.1](https://github.com/talaia-labs/python-teos/milestone/1) since it is a pretty trivial change that would make bootstrapping way easier.

PR #174, opened 2020-07-09
Move data folders in subfolders based on the network; automated key generation for first startup

bigspider:
As discussed in #169, this PR moves the datafolder in a subfolder for teosd based on the network for regtest or testnet (not for mainnet).
The data folder for mainnet is unchanged.

Moreover (thanks to @sr-gi), this also adds some key management features as described in #170, which also simplifies the steps for running the tower for the first time.

Closes: #169, #170


common/config_loader.py
```diff
@@ -110,9 +110,17 @@ def extend_paths(self):
...
 
+        # if there is a BTC_NETWORK parameter whose value is not "mainnet", we append it to the data_dir when expanding
+        network_field = self.conf_fields.get("BTC_NETWORK")
+        network = network_field["value"] if network_field else ""
+        network_suffix = "" if network == "mainnet" else network
+
         for key, field in self.conf_fields.items():
             if field.get("path") and isinstance(field.get("value"), str):
-                self.conf_fields[key]["value"] = os.path.join(self.data_dir, self.conf_fields[key]["value"])
+                base_path = os.path.join(self.data_dir, self.conf_fields[key]["value"])
```
sr-gi, line 125: base_path is not used

bigspider: Removed in 18fe3e8

PR #183, opened 2020-07-23
Remove bitcoind mock from unit tests

### LINKED ISSUE #104: 500 Internal Server Error on bitcoin-cli getblock
Every now and then we get a 500 Internal Server Error from bitcoind when asking for a block. This causes the software to crash since a notified block is supposed to always be there.

```
17/03/2020 12:16:35 [BlockProcessor]: Couldn't get block from bitcoind (error={'code': -342, 'message': "non-JSON HTTP response with '500 INTERNAL SERVER ERROR' from server"}).
Exception in thread Thread-965:
Traceback (most recent call last):
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/opt/circleci/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/circleci/repo/teos/watcher.py", line 143, in do_watch
    logger.info("New block received", block_hash=block_hash, prev_block_hash=block.get("previousblockhash"))
AttributeError: 'NoneType' object has no attribute 'get'
```

I haven't been able to properly reproduce the error.

We could work around it by checking if the received data is None, but the may makes us miss data. It will be better to find the root of the problem.

sr-gi:
Runs unit tests against bitcoind instead of bitcoind-mock.

The purpose of the mock was to make testing reorgs easier, since I thought testing it with bitcoind may have required running a dockerized regtest. Turns out that's not necessary, so tests are better run against bitcoind to avoid any inconsistencies. Furthermore, this saves us the time and effort to maintain bitcoind-mock.

A global `conftest.py` has been created for `teos` tests to remove some replicated methods and avoid replicating some more. On top of that, test now do run in temporary folders (also does bitcoind) to not mess up current data.

circle-ci has been updated to download `bitcoind` from bitcoincore.org instead of getting it from snap. That also allows us to cache the files and reduces the testing time. Currently it is running 0.20. Manual update may be required if some interesting features are released.

Finally, `test_malformed_penalty` has been updated since `bitcoin-20.0` uses `bech32` addresses by default, and that was making the test to fail.

Sort of unrelated change:

`DATA_DIR` has been added to the config object in `ConfigLoader.build_config`. Turns out the dir was not being properly rewritten if the param was passed to `teosd`. Noticed while configuring the tests to run in a temporary folder so I needed to add it to this PR.

Close #104


sr-gi:
Even though it's not passing integration tests, feel free to give it a look @bigspider. Currently it should only fail the last two `block_processor` tests, since those are the ones that test reorgs.

I think this would be a nice opportunity to give a close look to the tests and change anything that may have been carried out from old versions and may not be required anymore. It may also be a good time to add any missing comments on what some tests are doing if they are not properly explained already.

test/teos/unit/test_block_processor.py
```diff
@@ -13,9 +14,9 @@
 )
 
 
-def test_get_best_block_hash(run_bitcoind, block_processor):
+def test_get_best_block_hash(block_processor):
     best_block_hash = block_processor.get_best_block_hash()
-    # As long as bitcoind is running (or mocked in this case) we should always a block hash
+    # As long as bitcoind is running we should always a block hash
```
bigspider, line 19: typo "always a block hash"

test/teos/unit/test_block_processor.py
```diff
@@ -90,22 +87,18 @@ def test_is_block_in_best_chain(block_processor):
...
-
-    # Create a fork (forking creates a block if the mock is set by events)
-    fork(ancestor)
+    blocks = generate_blocks(3)
+    best_block_hash = blocks[-1]
 
-    # Create another block to make the best tip change (now both chains are at the same height)
-    generate_blocks(5)
+    # Create a fork (invalidate the next block after the ancestor amd mine 4 blocks on top)
+    fork(blocks[0], 4)
```
bigspider, line 101: typo "amd"

test/teos/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+        except JSONRPCException as e:
+            if "Loading wallet..." in str(e):
+                sleep(1)
+
+    # Mine enough blocks so coinbases are mature and we have enough funds to run everything
+    bitcoin_cli.generatetoaddress(105, btc_addr)
+    create_initial_transactions()
+
+
+def create_initial_transactions(fee=Decimal(5 / pow(10, 5))):
```
bigspider, line 87: Not a big deal, but the default `fee` above results in `Decimal('0.0000500000000000000023960868011929647991564706899225711822509765625')` due to floating point encoding issues; maybe cleaner to just write `fee=Decimal('0.00005')`.

test/teos/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+    if isinstance(commitment_txs, list):
+        for tx in commitment_txs:
+            bitcoin_cli.sendrawtransaction(tx)
+    elif isinstance(commitment_txs, str):
+        bitcoin_cli.sendrawtransaction(commitment_txs)
+
+    return generate_blocks(1)
+
+
+def create_commitment_tx(utxo=None, destination=None, fee=Decimal(1 / pow(10, 5))):
```
bigspider, line 149: Same as above for the fee.

test/teos/conftest.py
```diff
@@ -0,0 +1,199 @@
...
+        utxo = utxos.pop(0)
+
+    return utxo
+
+
+def generate_blocks(n):
+    return bitcoin_cli.generatetoaddress(n, btc_addr)
+
+
+def generate_blocks_w_delay(n):
```
bigspider, line 129: `_with_delay`?

test/teos/e2e/conftest.py
```diff
@@ -1,93 +1,30 @@
...
-
-from common.config_loader import ConfigLoader
 from common.cryptographer import Cryptographer
-
-getcontext().prec = 10
-utxos = []
+from test.teos.conftest import config
 
 
+# This fixture needs to be manully run on the first E2E.
```
bigspider, line 12: typo `manully`.

test/teos/unit/test_carrier.py
```diff
@@ -1,38 +1,42 @@
...
+from test.teos.unit.conftest import get_random_value_hex
 from teos.rpc_errors import RPC_VERIFY_ALREADY_IN_CHAIN, RPC_DESERIALIZATION_ERROR
 
+from test.teos.conftest import create_commitment_tx, bitcoin_cli
+
 
 # FIXME: This test do not fully cover the carrier since the simulator does not support every single error bitcoind may
 #        return for RPC_VERIFY_REJECTED and RPC_VERIFY_ERROR. Further development of the simulator / mocks or simulation
 #        with bitcoind is required
+# TODO: Given we are getting rid of the mock, test the uncovered cases.
```
bigspider, line 11: TODO here, maybe open a new issue for it?

test/teos/unit/test_block_processor.py
sr-gi: Fixed

test/teos/unit/test_block_processor.py
sr-gi: Fixed

test/teos/conftest.py
sr-gi: Fixed

test/teos/e2e/conftest.py
sr-gi: Fixed

test/teos/unit/test_carrier.py
sr-gi: Just opened one https://github.com/talaia-labs/python-teos/issues/184

test/teos/unit/test_block_processor.py
bigspider: Still missing the verb in the sentence?

sr-gi: I think that's showing the old one:

https://github.com/talaia-labs/python-teos/blob/a79ec735b7c441686eef6070dd9dba6c2e571700/test/teos/unit/test_block_processor.py#L19

bigspider: Oh, weird, I tried opening the file and still saw the old one.

Issue #166, opened 2020-07-02
Refactor tests so that they can be ran individually


bigspider:
Several of the test suites are written in such a way that tests will work if the whole suite is ran in order, but some tests might fail if ran individually, as they depend on side effects of previous tests.

It would be better to make each test independent, perhaps with wider use of fixtures to always set up everything needed for each test.

PR #188, opened 2020-08-03
Refactor tests so they can be ran individually

bigspider:
Make the dependency to the run_bitcoind fixture explicit in tests that need it, so that unit tests can be ran individually.

This PR only takes care of the unit tests, not the e2e tests.

After that, 6 tests needed fixing as they functionally depended on the side effects of previous tests in the same test module:
- `test_get_appointment_in_watcher` and `test_get_appointment_in_responder` in test_api.py
- `test_delete_locator_map`, `test_delete_watcher_appointment` and `test_delete_responder_tracker` in test_appointments_dbm.py
- `test_do_watch_cache_update` in test_watcher.py

Fixed all of the above, except `test_do_watch_cache_update`: that is basically an integration test (together with the previous test, `test_do_watch`), so it might perhaps be better refactored out of the unit tests suite, but that is out of scope for this ticket. Left a TODO there.
 
Closes: #166

### DIFF
--- a/test/teos/unit/test_api.py
+++ b/test/teos/unit/test_api.py
@@ -66,7 +66,7 @@ def get_all_db_manager():
     rmtree("get_all_tmp_db")
 
 
-@pytest.fixture(scope="module", autouse=True)
+@pytest.fixture(scope="module")
 def internal_api(run_bitcoind, db_manager, gatekeeper, carrier, block_processor):
     responder = Responder(db_manager, gatekeeper, carrier, block_processor)
     watcher = Watcher(
@@ -199,10 +199,7 @@ def test_add_appointment(internal_api, client, appointment, block_processor):
     assert r.json.get("start_block") == block_processor.get_block_count()
 
 
-def test_add_appointment_no_json(internal_api, client, appointment):
-    # Simulate the user registration (end time does not matter here)
-    internal_api.watcher.gatekeeper.registered_users[user_id] = UserInfo(available_slots=1, subscription_expiry=0)
-
+def test_add_appointment_no_json(client):
     # No JSON data
     r = client.post(add_appointment_endpoint, data="random_message")
     assert r.status_code == HTTP_BAD_REQUEST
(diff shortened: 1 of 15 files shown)


test/teos/unit/test_api.py
```diff
@@ -461,13 +461,20 @@ def test_get_appointment_not_registered_user(client):
...
-def test_get_appointment_in_watcher(api, client, appointment):
+def test_get_appointment_in_watcher(client, api, appointment, monkeypatch):
+
     # Mock the appointment in the Watcher
     uuid = hash_160("{}{}".format(appointment.locator, user_id))
     extended_appointment_summary = {"locator": appointment.locator, "user_id": user_id}
     api.watcher.appointments[uuid] = extended_appointment_summary
     api.watcher.db_manager.store_watcher_appointment(uuid, appointment.to_dict())
 
+    def mock_authenticate_user(*args, **kwargs):
```
sr-gi, line 472: Given this is defined twice, it could be a place before the tests as a non-internal method, just with the rest of the non-test functions

test/teos/unit/test_appointments_dbm.py
```diff
@@ -18,7 +18,7 @@
 
 
 @pytest.fixture(scope="module")
-def watcher_appointments():
+def watcher_appointments(run_bitcoind):
```
sr-gi, line 21: We should replace the `generate_dummy_tracker fixture` to call `generate_dummy_tracker`

test/teos/unit/test_api.py
bigspider: Done in 3d016331a1a94d4eade971a6f1955c4ff23294dc.

test/teos/unit/test_appointments_dbm.py
bigspider: Done in 3d016331a1a94d4eade971a6f1955c4ff23294dc, and done a heavier refactoring of both the functions (that are now fixtures) in bcea9473c5b7a34d74f1c2631f22d534f2922f8c.

Issue #191, opened 2020-08-05
Manage how the HTTP API is run


sr-gi:
Once #187 is merged, the HTTP API will be handled by `gunicorn` instead of using the non-production-ready flask webserver.

This is an improvements over the previous approach, but may not be good for all cases. We should give a bit of flexibility about how a user wants to run this. 

Add some config options that allow users to choose how they would like to run this.

Check https://github.com/talaia-labs/python-teos/pull/187#issuecomment-668697966 and https://github.com/talaia-labs/python-teos/pull/187#pullrequestreview-461452113

PR #196, opened 2020-08-07
Pick WSGI server to run the API

sr-gi:
`gunicorn` is used as the WSGI server to run the api by default, however, in other setups it may make sense using something different:

- Windows machines cannot run gunicorn, so `uWSGI` may be an alternative.
- For testing and for setups where only a few (trusted) users are using the tower, Flask may suffice.

This PR adds the config parameters needed to allow the customisation of the WSGI server and adapt the daemon to run it.

partially covers #191  (not covers `uWSGI` for Windows environments, will leave as a followup since I cannot tests it personally).

This also fixes the issue with `e2e` testing leaving the `gunicorn` running. End-to-end test switched back to use `flask` now there's an option for it.


teos/api.py
```diff
@@ -56,26 +56,30 @@ def get_request_data_json(request):
...
+    or by the Flask development server, in which case it is not.
 
     Args:
-        internal_api_endpoint (:obj:`str`): Endpoint where the internal api is running (host:port).
-        min_to_self_delay (:obj:`str`): The minimum to_self_delay accepted by the Inspector.
-        log_file (:obj:`str`): The file_path where to store logs.
+        internal_api_endpoint (:obj:`str`): endpoint where the internal api is running (host:port).
+        endpoint (:obj:`str`): endpoint where the http api will be running (host:port).
+        min_to_self_delay (:obj:`str`): the minimum to_self_delay accepted by the Inspector.
+        log_file (:obj:`str`): the file_path where to store logs (only necessary if not running with Flask).
```
bigspider, line 70: What happens if not running with flask, but `log_file` is `None`? Worth adding a check?

teos/teosd.py
```diff
@@ -227,28 +235,28 @@ def main(config):
...
+            "rpcport=",
+            "btcnetwork=",
+            "btcrpcuser=",
+            "btcrpcpassword=",
+            "btcrpcconnect=",
+            "btcrpcport=",
+            "btcfeedconnect=",
+            "btcfeedport=",
+            "datadir=",
+            "wsgi=" "daemon",
```
bigspider, line 254: missing comma

teos/teosd.py
```diff
@@ -198,16 +199,23 @@ def main(config):
...
+                    [
+                        "gunicorn",
+                        f"--bind={api_endpoint}",
+                        f"teos.api:serve(internal_api_endpoint='{INTERNAL_API_ENDPOINT}', "
+                        f"endpoint='{api_endpoint}', min_to_self_delay='{config.get('MIN_TO_SELF_DELAY')}', "
+                        f"log_file='{config.get('LOG_FILE')}')",
+                    ]
+                )
+            else:
+                os.environ["WERKZEUG_RUN_MAIN"] = "true"
```
bigspider, line 216: Is it necessary to suppress the warning? After all, it's giving good advice.

bigspider (changes requested):
No major comment, apart from a typo that breaks the --wsgi argument.

teos/api.py
sr-gi: I've changed the way this works in e7f8a437fdd766ef8744e92e425cb7df3622d60b and now the log file is required.

teos/teosd.py
sr-gi: Fixed in 54e7f24e79d8e9497fe543cfea839b6c7677453e

teos/teosd.py
sr-gi: Fixed in 54e7f24e79d8e9497fe543cfea839b6c7677453e

sr-gi:
This should be working now. Happened to find an issue regarding python 3.8 and MacOS that extended to Windows also. It should work with versions prior to 3.8 and 3.8 onwards with the current version.

bigspider (approved):

Issue #182, opened 2020-07-23
Add more features to the cli


bigspider:
Commands to add to the `cli` once it is split from the client:

- `gettowerinfo`: Get a summary of the data in the tower, such as tower_id, number of registered users, appointments in the watcher and responder, ...
- `getusers`: Gets a list of users
- `getuser "user_id"`: Get info about a given user
- `getappointment "locator"`: Gets info about all the appointments given a locator

_Originally posted by @sr-gi in https://github.com/talaia-labs/python-teos/issues/146#issuecomment-659260096_

PR #197, opened 2020-08-10
New cli features

bigspider:
Commands added to the internal api and to teos_cli:
    get_tower_info: Get a summary of the data in the tower, such as tower_id, number of registered users, appointments in the watcher and responder, ...
    get_users: Gets the list of all registered user ids
    get_user "user_id": Get info about a given user

Other related changes in this PR:
  - refactored teos_cli with an RPCClient class that has wrappers for the calls to the rpc service
  - added some accessors to the watcher's state, used by the internal api
  - removed test_rpc.py, moved existing get_all_appointment tests to the test_internal_api.py

Did not add a `get_appointments <locator>` command, as it requires a bigger amount of work in the watcher; opened a new issue #199.

Closes: #182


teos/internal_api.py
```diff
@@ -144,3 +147,71 @@ def get_all_appointments(self, request, context):
...
+    def get_appointments(self, request, context):
+        try:
+            Inspector.check_locator(request.locator)
+        except InspectionFailed as e:
+            context.set_details(f"Invalid locator: {e.reason}")
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            return GetAppointmentsResponse()
+
+        with self.rw_lock.gen_rlock():
+            uuids = self.watcher.locator_uuid_map.get(request.locator)
```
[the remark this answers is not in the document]
sr-gi: Also, this won't work straightaway. If the locator is not in the Watcher, it can be in the Responder.

I think there's no good way of getting al this atm, since if the appointment is in the Responder, there is no way of fetching it by locator. We may need a new map for that.

teos/cli/teos_cli.py
```diff
@@ -54,19 +73,55 @@ def main(command, args, command_line_conf):
...
+    rpc_client = RPCClient(teos_rpc_host, teos_rpc_port)
+
     try:
         if command == "get_all_appointments":
-            appointment_data = get_all_appointments(config.get("RPC_BIND"), config.get("RPC_PORT"))
-            if appointment_data:
-                print(json.dumps(appointment_data, indent=4, sort_keys=True))
+            result = rpc_client.get_all_appointments()
+
+        elif command == "get_appointments":
```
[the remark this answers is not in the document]
bigspider: Added in 3323ba38cfe670ac88830f6b29f941c6d9a382c3.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,55 @@
...
+    @formatted
+    def get_users(self):
+        return self.stub.get_users(Empty())
 
-    # FIXME: Handle different errors
-    except grpc.RpcError:
-        print("Can't connect to the Eye of Satoshi. RPC server cannot be reached", file=sys.stderr)
-        return None
+    @formatted
+    def get_user(user_id):
```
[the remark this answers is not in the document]
bigspider: Fixed in 3323ba38cfe670ac88830f6b29f941c6d9a382c3.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,55 @@
...
+from teos.protobuf.user_pb2 import GetUserRequest
 
 
-def get_all_appointments(rpc_host, rpc_port):
-    """
-    Gets information about all appointments stored in the tower.
+def formatted(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        result = func(*args, **kwargs)
```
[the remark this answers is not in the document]
bigspider: Now all the commands should work fine and the json pretty-printed.

teos/internal_api.py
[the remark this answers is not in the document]
bigspider: Mostly done in 5154047f2fa083a2bedca19a687c3780f44c1d44 and 9ea435a3086f5dc25c1e1b6a46e081a492b875c3. Removed the get_appointments cli command in 3d3c47c0c6d5a45ef4f56476e09514440afd0007, will be added as a separate ticket.

teos/cli/teos_cli.py
```diff
@@ -54,36 +83,66 @@ def main(command, args, command_line_conf):
...
+
                 else:
                     sys.exit("Unknown command. Use help to check the list of available commands")
 
             else:
                 sys.exit(show_usage())
 
+    except grpc.RpcError as e:
+        sys.exit(e.details())
     except (FileNotFoundError, IOError, ConnectionError, ValueError) as e:
```
sr-gi, line 132: I don't think most of this exceptions are thrown anymore. 

- `FileNotFoundError` may come from the split with the client when the tower key needed to be loaded.
- `ConnectionError` was for http connections to the API endpoint. I don't think this applies to gRPC. I'm pretty sure there are all included in `grpc. RpcError`, but that should be looked closely.
- `IOError` applied to both reading from the connection and the file, so it may not apply here anymore.

teos/cli/teos_cli.py
```diff
@@ -54,36 +83,66 @@ def main(command, args, command_line_conf):
...
+                elif command == "get_user":
+                    sys.exit(help_get_user())
+
                 else:
                     sys.exit("Unknown command. Use help to check the list of available commands")
 
             else:
                 sys.exit(show_usage())
 
+    except grpc.RpcError as e:
```
sr-gi, line 130: This may need better handling to provide a proper feedback to the user. Split the error based on the error code and provide some good reasons to the errors.

For example, if the tower is not online you'll get `failed to connect to all addresses` atm, which is not descriptive at all.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,65 @@
...
 from teos.protobuf.tower_services_pb2_grpc import TowerServicesStub
+from teos.protobuf.user_pb2 import GetUserRequest
 
 
-def get_all_appointments(rpc_host, rpc_port):
+def to_json(obj):
+    return json.dumps(obj, indent=4)
+
+
+def formatted(func):
```
sr-gi, line 33: This does not seem to fit here, specially given it could be reused for the client or other modules that need pretty data to be shown. `common.tools` may be a better fit.

Also, this has no checks at all, maybe make sure that at least the input is a string so it can be dumped as json. Also add docs.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,65 @@
...
+
+    return wrapper
+
 
-    Returns:
-        :obj:`dict` a dictionary containing all the appointments stored by the Responder and Watcher if the tower
-        responds.
+class RPCClient:
+    """Creates and keeps a connection to the an RPC serving TowerServices. It has methods to call each of the
+    available grpc services, and it returns a pretty-printed json response.
```
sr-gi, line 51: Arguments and Attributes are missing in the docs.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,65 @@
...
-            response = json_format.MessageToDict(
-                r.appointments, including_default_value_fields=True, preserving_proto_field_name=True
-            )
+    def __init__(self, rpc_host, rpc_port):
+        self.rpc_host = rpc_host
+        self.rpc_port = rpc_port
+        self.channel = grpc.insecure_channel(f"{rpc_host}:{rpc_port}")
+        self.stub = TowerServicesStub(self.channel)
+
+    @formatted
```
sr-gi, line 61: Missing docs for all the methods. A one liner would do.

teos/cli/teos_cli.py
```diff
@@ -14,37 +15,65 @@
...
-    except grpc.RpcError:
-        print("Can't connect to the Eye of Satoshi. RPC server cannot be reached", file=sys.stderr)
-        return None
+    def get_users(self):
+        result = self.stub.get_users(Empty())
+        return to_json(list(result.user_ids))
+
+    @formatted
+    def get_user(self, user_id):
+        return self.stub.get_user(GetUserRequest(user_id=user_id), indent=4)
```
sr-gi, line 76: I don't think that `indent=4` belongs there. Also, the return of this method looks as follows:

```
{
    "user": {
        "subscription_expiry": 4433.0,
        "appointments": {
            "421bc1a536a7e6c4d9138110b9421e54881c86ad": 1.0
        },
        "available_slots": 99.0
    }
}
```

I'd either get rid of the external dictionary (user: {}) or change it so intead of user it shows the id. e.g:

```
{
    0207213bde4b80da7e58bfff65cd0d3c4d7d349461595a534ea9949d2059d382ee: {
        "subscription_expiry": 4433.0,
        "appointments": {
            "421bc1a536a7e6c4d9138110b9421e54881c86ad": 1.0
        },
        "available_slots": 99.0
    }
}
```

I'd lean towards the first tho.

sr-gi: Also, the return contains the number of slots taken by every appointment:

"421bc1a536a7e6c4d9138110b9421e54881c86ad": **1.0**

I don't think that's relevant information.

teos/cli/teos_cli.py
```diff
@@ -54,36 +83,66 @@ def main(command, args, command_line_conf):
...
+        elif command == "get_users":
+            result = rpc_client.get_users()
+
+        elif command == "get_user":
+            if not args:
+                sys.exit("No user_id was given")
+            if len(args) > 1:
+                sys.exit(f"Expected only one argument, not {len(args)}")
 
+            result = rpc_client.get_user(args[0])
```
sr-gi, line 107: Add a sanity check to make sure the provided arguments is a compressed public key? (`common.tools.is_compressed_pk`)

teos/cli/teos_cli.py
```diff
@@ -54,36 +83,66 @@ def main(command, args, command_line_conf):
...
+    except grpc.RpcError as e:
+        sys.exit(e.details())
     except (FileNotFoundError, IOError, ConnectionError, ValueError) as e:
         sys.exit(str(e))
     except (InvalidKey, InvalidParameter, TowerResponseError, SignatureError) as e:
         sys.exit(f"{e.msg}. Error arguments: {e.kwargs}")
     except Exception as e:
         sys.exit(f"Unknown error occurred: {str(e)}")
 
+    if result:
```
sr-gi, line 139: Kinda nitpicking here, but would't this be better off before all the exceptions? In terms of readability basically. Also result may be not defined, but I don't think there is any care where the flow won't have been forced to exit before that.

teos/internal_api.py
```diff
@@ -137,10 +139,36 @@ def get_appointment(self, request, context):
...
-            responder_trackers = self.watcher.db_manager.load_responder_trackers()
+            watcher_appointments = self.watcher.get_all_watcher_appointments()
+            responder_trackers = self.watcher.get_all_responder_trackers()
 
         appointments = Struct()
         appointments.update({"watcher_appointments": watcher_appointments, "responder_trackers": responder_trackers})
 
         return GetAllAppointmentsResponse(appointments=appointments)
+
+    def get_tower_info(self, request, context):
```
sr-gi, line 150: Missing docs for all the methods. A one-liner would do (even if it points to the docs of another class)

test/teos/unit/test_internal_api.py
```diff
@@ -45,6 +50,17 @@ def internal_api(db_manager, gatekeeper, carrier, block_processor):
     i_api.rpc_server.stop(None)
 
 
+@pytest.fixture()
+def clear_state(internal_api, db_manager):
```
sr-gi, line 54: I'm guessing setting the db fixture to be cleared every test did not work?

test/teos/unit/test_internal_api.py
```diff
@@ -232,6 +248,139 @@ def test_get_appointment_non_existent(internal_api, stub):
...
+    internal_api.watcher.db_manager.store_watcher_appointment(uuid, appointment.to_dict())
+
+    response = stub.get_all_appointments(Empty())
+    appointments = dict(response.appointments)
+
+    assert len(appointments.get("watcher_appointments")) == 1 and len(appointments.get("responder_trackers")) == 0
+    assert dict(appointments.get("watcher_appointments")[uuid]) == appointment.to_dict()
+
+    # Delete the data
+    internal_api.watcher.db_manager.delete_watcher_appointment(uuid)
```
sr-gi, line 272: Is this necessary given the `clear_state` fixture?

test/teos/unit/test_internal_api.py
```diff
@@ -232,6 +248,139 @@ def test_get_appointment_non_existent(internal_api, stub):
...
+    internal_api.watcher.db_manager.store_responder_tracker(uuid, tracker.to_dict())
+
+    response = stub.get_all_appointments(Empty())
+    appointments = dict(response.appointments)
+
+    assert len(appointments.get("watcher_appointments")) == 0 and len(appointments.get("responder_trackers")) == 1
+    assert dict(appointments.get("responder_trackers")[uuid]) == tracker.to_dict()
+
+    # Delete the data
+    internal_api.watcher.db_manager.delete_responder_tracker(uuid)
```
sr-gi, line 289: Same as last comment.

test/teos/unit/test_internal_api.py
```diff
@@ -232,6 +248,139 @@ def test_get_appointment_non_existent(internal_api, stub):
...
+        {
+            "uid1": {"penalty_txid": "txid1", "locator": "locator1", "user_id": "user_id1"},
+            "uid2": {"penalty_txid": "txid2", "locator": "locator2", "user_id": "user_id2"},
+            "uid3": {"penalty_txid": "txid3", "locator": "locator2", "user_id": "user_id3"},
+        },
+    )
+
+    response = stub.get_tower_info(Empty())
+    assert isinstance(response, GetTowerInfoResponse)
+    assert response.tower_id == Cryptographer.get_compressed_pk(internal_api.watcher.signing_key.public_key)
```
sr-gi, line 342: There's a teos_pk defined globally in this test file.

test/teos/unit/test_internal_api.py
```diff
@@ -232,6 +248,139 @@ def test_get_appointment_non_existent(internal_api, stub):
...
+    assert isinstance(response, GetUserResponse)
+
+    assert json_format.MessageToDict(response.user) == mock_user_info.to_dict()
+
+
+def test_get_user_not_found(internal_api, stub):
+    mock_user_id = "some_non_existing_user_id"
+
+    with pytest.raises(grpc.RpcError) as e:
+        response = stub.get_user(GetUserRequest(user_id=mock_user_id))
```
sr-gi, line 383: No need to store the call in a variable.

test/teos/unit/test_internal_api.py
```diff
@@ -232,6 +248,139 @@ def test_get_appointment_non_existent(internal_api, stub):
...
+def test_get_user(internal_api, stub, monkeypatch):
+    # it doesn't matter they are not valid user ids and user data object for this test
+    mock_user_id = "some_user_id"
+    mock_user_info = UserInfo(100, 1234)
+
+    def mock_get_user_info(user_id):
+        if user_id == mock_user_id:
+            return mock_user_info
+        else:
+            raise f"called with an unexpected user_id: {user_id}"
```
sr-gi, line 369: This is wrong. You cannot raise a string, it would crash on runtime if reached.

test/teos/unit/test_internal_api.py
bigspider: I didn't try that, but I thought that it would be better to explicitly mark tests that rely on empty db.

teos/cli/teos_cli.py
bigspider: Removed all, there's indeed nothing that can throw errors other than the rpc calls. Left only `grpc.RpcError` and the generic `Exception` which should currently never happen. 8a8a2aae2ef3f1ddc05d732f9f9f96764447d0eb

teos/cli/teos_cli.py
bigspider: I added a tailored message for status code UNAVAILABLE, but decided against further splitting by status code, as in all the other cases that I know of the `details()` call should provide an appropriate description, as we generate them in the internal api. 8a8a2aae2ef3f1ddc05d732f9f9f96764447d0eb

teos/cli/teos_cli.py
bigspider: Re-added `InvalidParameter` in 8171b65.

teos/cli/teos_cli.py
bigspider: I was just dumping the whole UserInfo object; instead, now I'm explicitly taking the `subscription_expiry` and `available_slots` fields, and only putting the keys of the "appointments" dictionary as an array, so it looks like this after flattening:

```
{
    "subscription_expiry": 4320.0,
    "appointments": [],
    "available_slots": 100.0
}
```

(don't the floats, will be taken care of)

Done in 8171b65ea7350877a911b0f95f9a95cae00f1441

teos/cli/teos_cli.py
bigspider: Added in 8171b65ea7350877a911b0f95f9a95cae00f1441.

teos/cli/teos_cli.py
bigspider: I moved that inside the `try` block, I agree it's more readable. The reason I put the `if` is that in case we later have a command that can return empty output (but not because of failure) I didn't want to print an empty line.

teos/cli/teos_cli.py
bigspider: Added some docs in 7d75df8.

teos/internal_api.py
bigspider: Added docs in 7d75df8. As the methods are basically service handlers, I only put one-liners on what it's doing, without details on the specific request format and response.

teos/cli/teos_cli.py
bigspider: Added in 7d75df813325e0ce08717101ac10c3ec23e43862.

teos/cli/teos_cli.py
bigspider: I'm not sure this is (or can be made) generic enough to put it in a common module: the wrapped function must return something that can be passed to `json_format.MessageToDict` (e.g. e whole Message response, or some subfields), and the output dict is then passed to `to_json`, which is just a custom wrapper for `json.dumps` to make sure that the conversion to json is done in the same way for other places that don't user `formatted` (currently only `RPCClient.get_users`).

test/teos/unit/test_internal_api.py
bigspider: Removed in 7e8482d915bd416d957abeda3a0d0e8746bc5e94.

test/teos/unit/test_internal_api.py
bigspider: Removed in 7e8482d915bd416d957abeda3a0d0e8746bc5e94.

test/teos/unit/test_internal_api.py
bigspider: Done in 7e8482d915bd416d957abeda3a0d0e8746bc5e94.

test/teos/unit/test_internal_api.py
bigspider: Done in 7e8482d915bd416d957abeda3a0d0e8746bc5e94.

teos/cli/teos_cli.py
bigspider: Ok, added a comment in 4e14ad64b379327f65fc67724f18f182777e2dfb to clarify the purpose of to_json.

common/tools.py
```diff
@@ -77,3 +77,23 @@ def setup_data_folder(data_folder):
     """
 
     Path(data_folder).mkdir(parents=True, exist_ok=True)
+
+
+def intify(obj):
+    """Takes an object that is a recursive composition of primitive types, lists and dictionaries, and returns an
```
sr-gi, line 83: Some nits:

Formatting within docs needs double quotes.

Also, given the docs are not a oneliner, break after """ (for both opening and close):
```
"""
Docs
"""

code
```

test/common/unit/test_tools.py
```diff
@@ -102,3 +110,68 @@ def test_setup_data_folder():
...
+    assert type(res["c"]["d"][0] == int)
+
+
+def test_intify_changed():
+    assert intify(1.0) == 1
+    assert intify(-1.0) == -1
+
+    x = [1.0, 1.5, -2.0, True]
+    res = intify(x)
+    assert res == [1.0, 1.5, -2, True]
```
sr-gi, line 154: Shouldn't this be `[1, 1.5, -2, True]`?

teos/cli/teos_cli.py
```diff
@@ -54,36 +106,67 @@ def main(command, args, command_line_conf):
...
-    except (FileNotFoundError, IOError, ConnectionError, ValueError) as e:
-        sys.exit(str(e))
-    except (InvalidKey, InvalidParameter, TowerResponseError, SignatureError) as e:
-        sys.exit(f"{e.msg}. Error arguments: {e.kwargs}")
+        if result:
+            print(result)
+
+    except grpc.RpcError as e:
+        if e.code() == grpc.StatusCode.UNAVAILABLE:
+            sys.exit("It was not possible to reach the Eye of Satoshi. Check your network settings.")
```
sr-gi, line 158: I'd go with something more on the lines of "is the tower running", instead of check your network settings, specially since currently the cli works has no auth, so better to be used from localhost.

test/common/unit/test_tools.py
bigspider: Yup, although python doesn't check the types. Changed for consistency in 8909d91744bc8298f91aaca2eee48db72ee5e3d4.

common/tools.py
bigspider: Fixed in 8909d91744bc8298f91aaca2eee48db72ee5e3d4, also in several other files.

teos/cli/teos_cli.py
bigspider: Changed in 8236a9cd754d74c618aed817765ebd11a9997d6a.

PR #198, opened 2020-08-13
Fixes grpc with multiprocess

sr-gi:
Turns out gRPC in Python has some issues when using multiprocess. A way or working around it is to create the processes (fork) before instantiating anything related to gRPC.

This PR modifies the `InternalAPI` to be run in a similar way as the `RPC` does. Then the former is run in the main process but after the latter has been forked into a different process.

How the processes are handled have not changed from before (the main process takes care of the backend while the forked process takes care of the external RPC server).

Some info and related issues:

https://github.com/grpc/grpc/tree/b8241addc380e63b28be850c3da2fb04dc75212b/examples/python/multiprocessing

https://github.com/grpc/grpc/blob/master/doc/fork_support.md

https://github.com/grpc/grpc/issues/16001

### DIFF
--- a/teos/internal_api.py
+++ b/teos/internal_api.py
@@ -46,6 +46,26 @@ class InternalAPI:
         add_TowerServicesServicer_to_server(_InternalAPI(watcher, self.logger), self.rpc_server)
 
 
+def serve(watcher, internal_api_endpoint):
+    """
+    Serves the internal API at a given endpoint.
+
+    This method will serve and hold until the main process is started or a stop signal is received. Notice the later is
+    not possible possible currently since the stop signal has to be passed to `rpc_server` and it is not returned. This
+    may change once the stop command for the CLI is implemented.
+
+    Args:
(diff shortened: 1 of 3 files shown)


teos/internal_api.py
```diff
@@ -46,6 +46,26 @@ def __init__(self, watcher, internal_api_endpoint):
         add_TowerServicesServicer_to_server(_InternalAPI(watcher, self.logger), self.rpc_server)
 
 
+def serve(watcher, internal_api_endpoint):
+    """
+    Serves the internal API at a given endpoint.
+
+    This method will serve and hold until the main process is started or a stop signal is received. Notice the later is
+    not possible possible currently since the stop signal has to be passed to `rpc_server` and it is not returned. This
+    may change once the stop command for the CLI is implemented.
```
bigspider, line 55: Typo "latter" instead of "later"; also "possible" is repeated.

Issue #92, opened 2020-02-10
Block data during bootstrap from db


sr-gi:
When bootstrapping from db data the Watcher and Responder have to, potentially, update their states. If the process takes long enough, it could be the case that a block is received and missed, since the `ChainMonitor` is not up and running (to avoid messing with the old and new blocks).

A solution for this is needed so nothing is missed.

Discussion here would be appreciated, no need for an implemented solution to mark this challenge as completed

sr-gi:
Discussion here would be appreciated, no need for an implemented solution to mark this challenge as completed

PR #206, opened 2020-09-01
Improve ChainMonitor to avoid missing blocks during bootstrap

bigspider:
Added a multi-state life-cycle to the ChainMonitor. Also generalized to accept an arbitrary number of subscribers, instead of the two that we're currently using.

It is IDLE when created, it listens but does not notifies subscribers when LISTENING, it listens and notifies subscribers when ACTIVE, it stop any actions when TERMINATED.

Also improved the shutdown code to properly wait for a graceful termination of Watcher and Responder threads.

Closes: #92

### DIFF
--- a/teos/chain_monitor.py
+++ b/teos/chain_monitor.py
@@ -16,18 +16,18 @@ class ChainMonitorStatus(Enum):
 
 class ChainMonitor:
     """
-    The :class:`ChainMonitor` is in charge of monitoring the blockchain (via ``bitcoind``) to detect new blocks on top
+    The ``ChainMonitor`` is in charge of monitoring the blockchain (via ``bitcoind``) to detect new blocks on top
     of the best chain. If a new best block is spotted, the chain monitor will notify the given ``Queues``.
 
-    The :class:`ChainMonitor` monitors the chain using two methods: ``zmq`` and ``polling``. Blocks are only notified
+    The ``ChainMonitor`` monitors the chain using two methods: ``zmq`` and ``polling``. Blocks are only notified
     once per queue and the notification is triggered by the method that detects the block faster.
 
-    The :class:`ChainMonitor` lifecycle goes through 4 states: idle, listening, active and terminated.
-    When a :class:`ChainMonitor` instance is created, it is not yet monitoring the chain and the ``status`` attribute
+    The ``ChainMonitor`` lifecycle goes through 4 states: idle, listening, active and terminated.
+    When a ``ChainMonitor`` instance is created, it is not yet monitoring the chain and the ``status`` attribute
     is set to `ChainMonitorStatus.IDLE`.
     Once the ``monitor_chain`` method is called, the chain monitor changes ``status`` to
     `ChainMonitorStatus.LISTENING`, and starts monitoring the chain for new blocks; it does not yet notify the
-    receiving queues, but keeps the block hashes in the order they where spotted in an interna queue.
+    receiving queues, but keeps the block hashes in the order they where spotted in an internal queue.
     Once the ``activate`` method is called, the ``status`` changes to ``ChainMonitorStatus.ACTIVE``, and the receiving
     queues are notified in order for all the block hashes that are in the internal queue or any new one that is
     detected.
@@ -45,7 +45,7 @@ class ChainMonitor:
         logger: the logger for this component.
         best_tip (:obj:`str`): a block hash representing the current best tip.
         last_tips (:obj:`list`): a list of last chain tips. Used as a sliding window to avoid notifying about old tips.
-        terminate (:obj:`bool`): a flag to signal the termination of the :class:`ChainMonitor` (shutdown the tower).
+        terminate (:obj:`bool`): a flag to signal the termination of the ``ChainMonitor`` (shutdown the tower).
         check_tip (:obj:`Event`): an event that is triggered at fixed time intervals and controls the polling thread.
         lock (:obj:`Condition`): a lock used to protect concurrent access to the queues and ``best_tip`` by the zmq and
             polling threads.
@@ -122,14 +122,12 @@ class ChainMonitor:
         while self.status != ChainMonitorStatus.TERMINATED:
             self.check_tip.wait(timeout=self.polling_delta)
 
-            # Terminate could have been set while the thread was blocked in wait
-            if self.status != ChainMonitorStatus.TERMINATED:
-                current_tip = self.block_processor.get_best_block_hash()
+            current_tip = self.block_processor.get_best_block_hash()
 
-                # get_best_block_hash may return None if the RPC times out.
-                if current_tip and current_tip != self.best_tip and current_tip not in self.last_tips:
-                    self.logger.info("New block received via polling", block_hash=current_tip)
-                    self.enqueue(current_tip)
+            # get_best_block_hash may return None if the RPC times out.
+            if current_tip and current_tip != self.best_tip and current_tip not in self.last_tips:
+                self.logger.info("New block received via polling", block_hash=current_tip)
+                self.enqueue(current_tip)
 
     def monitor_chain_zmq(self):
         """
@@ -141,18 +139,16 @@ class ChainMonitor:
         while self.status != ChainMonitorStatus.TERMINATED:
             msg = self.zmqSubSocket.recv_multipart()
 
-            # Terminate could have been set while the thread was blocked in recv
-            if self.status != ChainMonitorStatus.TERMINATED:
-                topic = msg[0]
-                body = msg[1]
+            topic = msg[0]
+            body = msg[1]
 
-                if topic == b"hashblock":
-                    block_hash = binascii.hexlify(body).decode("utf-8")
-                    if block_hash != self.best_tip and block_hash not in self.last_tips:
-                        self.logger.info("New block received via zmq", block_hash=block_hash)
-                        self.enqueue(block_hash)
+            if topic == b"hashblock":
+                block_hash = binascii.hexlify(body).decode("utf-8")
+                if block_hash != self.best_tip and block_hash not in self.last_tips:
+                    self.logger.info("New block received via zmq", block_hash=block_hash)
+                    self.enqueue(block_hash)
 
-    def notify_listeners(self):
+    def notify_subscribers(self):
         """
         Once the method is fired, it keeps getting the elements added to the internal queue and notifies the receiving
         queues about them. It terminates whenever the internal state is set to ``ChainMonitorStatus.TERMINATED``.
@@ -170,7 +166,7 @@ class ChainMonitor:
 
     def monitor_chain(self):
         """
-        Changes the ``status`` of the :class:`ChainMonitor` from idle to listening. It initializes the ``best_tip`` to
+        Changes the ``status`` of the ``ChainMonitor`` from idle to listening. It initializes the ``best_tip`` to
         the current one (by querying the :obj:`BlockProcessor <teos.block_processor.BlockProcessor>`) and creates two
         threads, one per each monitoring approach (``zmq`` and ``polling``).
 
@@ -189,8 +185,8 @@ class ChainMonitor:
 
     def activate(self):
         """
-        Changes the ``status`` of the :class:`ChainMonitor` from listening to active. It creates a new thread that runs
-        the ``notify_listener`` method, which is in charge of notifying the receiving queue for each block hash that is
+        Changes the ``status`` of the ``ChainMonitor`` from listening to active. It creates a new thread that runs
+        the ``notify_subscribers`` method, which is in charge of notifying the receiving queue for each block hash that is
         added to the internal queue.
 
         Raises:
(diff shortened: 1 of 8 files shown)


test/teos/unit/test_responder.py
```diff
@@ -28,7 +28,7 @@
 @pytest.fixture(scope="module")
 def responder(db_manager, gatekeeper, carrier, block_processor):
     responder = Responder(db_manager, gatekeeper, carrier, block_processor)
-    chain_monitor = ChainMonitor(Queue(), responder.block_queue, block_processor, bitcoind_feed_params)
+    chain_monitor = ChainMonitor([Queue(), responder.block_queue], block_processor, bitcoind_feed_params)
```
sr-gi, line 31: The fixture is not being activated. This is OK since the do_watch method is the only one that requires chain monitoring. However, this made me realise that, most likely, all the other tests could use the fixture.

Mind giving it a try? It may imply slightly modifying some of the tests (like monkey patching some of the state)

sr-gi: PS: With monkey patching it may even work with only the fixture, since it does for the Watcher.

test/teos/unit/test_chain_monitor.py
```diff
@@ -2,8 +2,10 @@
 import time
 from queue import Queue
 from threading import Thread, Event, Condition
+import pytest
+import signal
```
sr-gi, line 6: Looks like signal is not used

test/teos/unit/test_chain_monitor.py
```diff
@@ -13,148 +15,245 @@ def test_init(block_processor):
...
+    assert chain_monitor.status == ChainMonitorStatus.IDLE
     assert chain_monitor.best_tip is None
     assert isinstance(chain_monitor.last_tips, list) and len(chain_monitor.last_tips) == 0
-    assert chain_monitor.terminate is False
+    assert chain_monitor.status == ChainMonitorStatus.IDLE
     assert isinstance(chain_monitor.check_tip, Event)
     assert isinstance(chain_monitor.lock, Condition)
     assert isinstance(chain_monitor.zmqSubSocket, zmq.Socket)
 
     # The Queues and asleep flags are initialized when attaching the corresponding subscriber
```
sr-gi, line 28: This comment needs an update

test/teos/unit/test_chain_monitor.py
```diff
@@ -13,148 +15,245 @@ def test_init(block_processor):
...
+    notifying_thread = Thread(target=chain_monitor.notify_listeners, daemon=True)
+    notifying_thread.start()
+
+    # the existing elements should be processed soon and in order for all queues
+    for q in [queue1, queue2]:
+        assert q.get(timeout=0.1) == block1
+        assert q.get(timeout=0.1) == block2
 
-    chain_monitor.notify_subscribers(new_block)
+    # Subscribers are only notified as long as they are awake
```
sr-gi, line 58: Since you're checking this, it may be worth checking that the queues are empty before starting the thread.

test/teos/unit/test_chain_monitor.py
```diff
@@ -13,148 +15,245 @@ def test_init(block_processor):
...
     # We can also force an update and see that it won't go through
-    assert chain_monitor.update_state(watcher_block) is False
+    assert chain_monitor.enqueue(queue0_block) is False
+
+    chain_monitor.terminate()
+    # The zmq thread needs a block generation to release from the recv method.
+    generate_blocks(1)
+
+
+def test_terminate(block_processor):
```
sr-gi, line 241: It'll be nice to generate blocks and check that they are not received.

test/teos/unit/test_chain_monitor.py
```diff
@@ -13,148 +15,245 @@ def test_init(block_processor):
...
+
+
+@pytest.mark.timeout(5)
+def test_threads_stop_when_terminated(block_processor):
+    # When status is "terminated", the methods running the threads should stop immediately
+
+    chain_monitor = ChainMonitor([Queue(), Queue()], block_processor, bitcoind_feed_params)
+    chain_monitor.terminate()
+
+    # If any of the function does not exit immrdiately, the test will timeout
```
sr-gi, line 256: Typo immrdiately

sr-gi (changes requested):
Looks pretty good, mainly minor comments.

A test I would like to be added would be the daemon being initialised with a backlog of data to update (`Builder`) and block being generated at the same time. Then check that once the Watcher and the Responder are ready (queues are populated with the old data) and the ChainMonitor is set to active, both Watcher and Responder process the data properly (no reorg complain).

teos/chain_monitor.py
```diff
@@ -59,38 +78,29 @@ def __init__(self, watcher_queue, responder_queue, block_processor, bitcoind_fee
...
-
-        Args:
-            block_hash (:obj:`str`): the new block hash to be sent to the subscribers.
-        """
-
-        self.watcher_queue.put(block_hash)
-        self.responder_queue.put(block_hash)
-
-    def update_state(self, block_hash):
+    def enqueue(self, block_hash):
```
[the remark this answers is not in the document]
bigspider: Tbh I don't think it hurts to have it (it was like this also before).
Checking the queue length after enqueuing might be unreliable if the thread that empties the queue is active, as it might remove an element before the check.

test/teos/unit/test_chain_monitor.py
bigspider: Fixed in cc60fa2.

test/teos/unit/test_chain_monitor.py
bigspider: Added in cc60fa2.

test/teos/unit/test_chain_monitor.py
bigspider: Removed the comment in cc60fa2.

test/teos/unit/test_chain_monitor.py
bigspider: Removed in cc60fa2.

teos/chain_monitor.py
```diff
@@ -1,22 +1,43 @@
...
 
     The :class:`ChainMonitor` monitors the chain using two methods: ``zmq`` and ``polling``. Blocks are only notified
     once per queue and the notification is triggered by the method that detects the block faster.
 
+    The :class:`ChainMonitor` lifecycle goes through 4 states: idle, listening, active and terminated.
+    When a :class:`ChainMonitor` instance is created, it is not yet monitoring the chain and the ``status`` attribute
+    is set to `ChainMonitorStatus.IDLE`.
+    Once the ``monitor_chain`` method is called, the chain monitor changes ``status`` to
+    `ChainMonitorStatus.LISTENING`, and starts monitoring the chain for new blocks; it does not yet notify the
+    receiving queues, but keeps the block hashes in the order they where spotted in an interna queue.
```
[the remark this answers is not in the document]
bigspider: Fixed in cc60fa2.

teos/chain_monitor.py
```diff
@@ -104,57 +114,100 @@ def update_state(self, block_hash):
...
+        If a new best tip is found, it is added to the internal queue.
         """
 
-        while not self.terminate:
+        while self.status != ChainMonitorStatus.TERMINATED:
             self.check_tip.wait(timeout=self.polling_delta)
 
             # Terminate could have been set while the thread was blocked in wait
-            if not self.terminate:
+            if self.status != ChainMonitorStatus.TERMINATED:
```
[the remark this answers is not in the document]
bigspider: Removed extra check in cc60fa2.

teos/chain_monitor.py
```diff
@@ -1,22 +1,43 @@
...
     """
     The :class:`ChainMonitor` is in charge of monitoring the blockchain (via ``bitcoind``) to detect new blocks on top
-    of the best chain. If a new best block is spotted, the chain monitor will notify the
-    :obj:`Watcher <teos.watcher.Watcher>` and the :obj:`Responder <teos.responder.Responder>` using ``Queues``.
+    of the best chain. If a new best block is spotted, the chain monitor will notify the given ``Queues``.
 
     The :class:`ChainMonitor` monitors the chain using two methods: ``zmq`` and ``polling``. Blocks are only notified
     once per queue and the notification is triggered by the method that detects the block faster.
 
+    The :class:`ChainMonitor` lifecycle goes through 4 states: idle, listening, active and terminated.
```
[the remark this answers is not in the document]
bigspider: Should be done, although I'm still not 100% sure when we're using single and when dobule quotes. Is there a place where I can read that?

teos/chain_monitor.py
```diff
@@ -59,38 +78,29 @@ def __init__(self, watcher_queue, responder_queue, block_processor, bitcoind_fee
...
         Args:
-            block_hash (:obj:`block_hash`): the new best tip.
+            block_hash (:obj:`str`): the new best tip.
 
         Returns:
-            :obj:`bool`: True is the state was successfully updated, False otherwise.
+            :obj:`bool`: True if the state was successfully updated, False otherwise.
         """
 
         if block_hash != self.best_tip and block_hash not in self.last_tips:
```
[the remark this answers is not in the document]
bigspider: I removed the `best_tip` attribute in 48903ef, adapting tests to check `last_tips` instead where needed.

teos/chain_monitor.py
```diff
@@ -104,57 +114,100 @@ def update_state(self, block_hash):
...
-                    self.lock.release()
+                        self.enqueue(block_hash)
+
+    def notify_listeners(self):
+        """
+        Once the method is fired, it keeps getting the elements added to the internal queue and notifies the receiving
+        queues about them. It terminates whenever the internal state is set to ``ChainMonitorStatus.TERMINATED``.
+        """
+
+        while self.status != ChainMonitorStatus.TERMINATED:
```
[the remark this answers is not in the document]
bigspider: Added the special END message in 9136508672d24b693eaa92d6fc4b665b3ecd7402.

Didn't add anything to make sure that the queue is emptied, and tbh I don't think we should, as normally the queue is expected to be almost always empty anyway (that is: new block comes, it's processed, waaaaaaait)

teos/teosd.py
```diff
@@ -298,7 +300,7 @@ def teardown(self):
         logger.info("Internal API stopped")
 
         # terminate the ChainMonitor
-        self.chain_monitor.terminate = True
+        self.chain_monitor.terminate()
```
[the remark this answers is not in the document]
bigspider: Refactored as we discussed in 460e6be. Now The "END" message of the ChainMonitor is propagated to the subscribers (watcher and responder) and used to terminate the `do_watch` threads, and TeosDaemon joins those threads before closing the db.

teos/chain_monitor.py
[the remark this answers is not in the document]
bigspider: Further refactored after 460e6be.

test/teos/unit/test_responder.py
bigspider: Hopefully done in d124cc07560c6b209a068dcf79e44a813c5cc614.

Not sure if this is what you meant, but I changed the `responder` fixture to be per-test (so it's a clean responder every time) and added setup and teardown, and changed all the tests to use the fixture instead of creating their responder (slight difference to how tests were done before: awaking the responder and activating the chain_monitor is now part of the fixture).

Couldn't do it only in `test_do_watch` as the test seems to need more control on the creation and starting time.

test/teos/unit/test_chain_monitor.py
bigspider: Done in dc09ef0.

teos/teosd.py
```diff
@@ -218,9 +221,8 @@ def bootstrap_components(self):
             elif len(missed_blocks_responder) != 0 and len(missed_blocks_watcher) != 0:
                 Builder.update_states(self.watcher, missed_blocks_watcher, missed_blocks_responder)
 
-        # Fire ChainMonitor
-        # FIXME: 92-block-data-during-bootstrap-db
-        self.chain_monitor.monitor_chain()
+        # Activate ChainMonitor
+        self.chain_monitor.activate()
```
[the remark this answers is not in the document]
bigspider: As commented in chat, I planned `start_services` more as bringing up all the interfaces, while `bootstrap_components` for starting everything internal of the watchtower. So leaving the way it is for now − it might need to be further refactored at some point to make the TeosDaemon testable, I suppose, but leaving that for future PRs.

teos/chain_monitor.py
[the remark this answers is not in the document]
sr-gi: I think this may be the best place https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

The rule of thumb is: double quote variable names you'd like to highlight, single quotes only work within class / object references (e.g: :obj:`Name <path>`)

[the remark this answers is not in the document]
sr-gi: There are still some btw:

L27: ChainMonitorStatus.IDLE
L29: ChainMonitorStatus.LISTENING
L53-54:  IDLE, LISTENING, ACTIVE and TERMINATED

teos/chain_monitor.py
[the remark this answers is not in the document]
sr-gi: OK, fair enough

teos/chain_monitor.py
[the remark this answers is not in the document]
bigspider: Fixed in 3dabf51.

test/teos/unit/test_responder.py
sr-gi: I may have to go over both Watchers and Responders test to simplify this further, so for now its OK.

sr-gi (approved):
LGTM

Issue #207, opened 2020-09-02
Make logging multiprocess-safe


bigspider:
Since we switched to a multiprocess architecture, we're using logging in an unsafe way: the `logging` package is, in fact, multithread-safe but not multiprocess-safe, see [the logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) on this.

PR #213, opened 2020-09-09
Added a tcp logging server that all other processes can stream logs to

bigspider:
In order to avoid having multiple processes writing to the same log file, which is not safe with python's logging system, this PR creates a tcp service in a process spawned by the main process, served on port 9020; all the other processes can then stream log requests to this logging server.

As the client no longer uses the internal logger, `logger.py` was moved to `/teos` from `/common`.

Also includes some unrelated small improvements to tests.

Closes: #207


teos/logging_server.py
```diff
@@ -0,0 +1,150 @@
...
+    Simple TCP socket-based logging receiver.
+    """
+
+    allow_reuse_address = True
+
+    def __init__(self, host="localhost", port=TCP_LOGGING_PORT, handler=LogRecordStreamHandler):
+        socketserver.ThreadingTCPServer.__init__(self, (host, port), handler)
+        self.abort = 0
+        self.timeout = 1
+        self.logname = None
```
sr-gi, line 108: I don't think logname is being used.

teos/logging_server.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def serve_until_stopped(self):
+        abort = 0
+        while not abort:
```
sr-gi, line 112: My understanding is that this is stopped when self.abort is set to `1` after a new log is read, given that if that does not happen abort is always `0`.

I think this could be better handled with a signal in the same way we're handling other processes.

sr-gi (changes requested):
Looks good overall, there are only a few small improvements that can be done:

- `common.logger` is not being used by client anymore (nor it will be), so it can be moved to `teos`. 
- Given `logger` will be part of `teos`, `logging_server` can be merged into `logger`.

There are also some minor improvements for `logging_server` that can be added, I'll share a diff.

teos/logging_server.py
bigspider: I tried with signals, but didn't quite work; instead, to keep things simple (the logging process is the last service that should stop anyway), I changed `serve_until_stopped` to `serve_forever` in 482b429, and removed the `abort` flag from `LogRecordSocketReceiver`.

teos/logging_server.py
bigspider: Removed in 482b429.

teos/logging_server.py
sr-gi: Since you're overwriting the base class method, stick to the method signature:

```
def serve_forever(self, poll_interval=0.5)
```

sr-gi:
Docs need to be updated to match the new path of `Logger`:

:obj:`Logger <common.logger.Logger>` -> :obj:`Logger <teos.logger.Logger>`

teos/logging_server.py
bigspider: Oops, I didn't realize it was overwriting a method. Fixed.

contrib/client/help.py
```diff
@@ -10,7 +10,7 @@ def show_usage():
         "\n\nGLOBAL OPTIONS:"
         "\n\t--apiconnect \tAPI server where to send the requests. Defaults to 'localhost' (modifiable in conf file)."
         "\n\t--apiport \tAPI port where to send the requests. Defaults to '9814' (modifiable in conf file)."
-        "\n\t-d, --debug \tShows debug information and stores it in teos_client.log."
+        "\n\t-d, --debug \tShows debug information and stores."
```
bigspider, line 13: En passant, so I don't forget: should remove `" and stores"`.

sr-gi: I'm already working on a miscellaneous PR including that. Will push in a bit.

teos/logger.py
```diff
@@ -0,0 +1,225 @@
...
+        logger = logging.getLogger(record.name)
+        logger.log(record.levelno, message, exc_info=record.exc_info, stack_info=record.stack_info)
+
+
+class LogRecordSocketReceiver(socketserver.ThreadingTCPServer):
+    """
+    Simple TCP socket-based logging receiver.
+    """
+
+    allow_reuse_address = True
```
sr-gi, line 184: Is this ever used?

bigspider: I think it needs to be in the class, see for example  https://stackoverflow.com/questions/15260558/python-tcpserver-address-already-in-use-but-i-close-the-server-and-i-use-allow

teos/rpc.py
```diff
@@ -67,7 +67,7 @@ class _RPC(TowerServicesServicer):
 
     Args:
         internal_api_endpoint (:obj:`str`): the endpoint where to reach the internal (gRPC) api.
-        logger (:obj:`Logger <common.logger.Logger>`): the logger for this component.
+        logger (:obj:`Logger <teos.logger.Logger>`): the logger for this component.
```
sr-gi, line 70: `log_file` is not used anymore in `serve`

teos/logger.py
sr-gi: Interesting.

sr-gi (changes requested):
LGTM.

Update PR description and remove `log_file` from `logger.serve` (but not from the docs, I have that in a followup)

bigspider:
> LGTM.
> 
> Update PR description and remove `log_file` from `logger.serve` (but not from the docs, I have that in a followup)

I think you were referring to rpc.serve; done in edebdc5.

sr-gi (approved):

Issue #193, opened 2020-08-06
Simplify gRPC stubs (if possible)


sr-gi:
I'm not completely sure about this, but leaving a note just in case.

Currently the gRPC stubs are created to be short-lived (a stub is created for every gRPC request and closed afterwards). Code could be simplified if the stub was created as part of some of the classes (for instance the HTTP API could have a long-lived stub with the InternalAPI).

PR #219, opened 2020-09-14
Modified API to reuse the grpc stub

bigspider:
It looks like reusing the channel/stub is the correct way of making repeated grpc requests, as channels already support multiplexing concurrent requests on them.

This PR recycles the `stub` for the `API` class, and adds missing attribute docs for `_RPC`.

Closes: #193

### DIFF
--- a/teos/api.py
+++ b/teos/api.py
@@ -99,6 +99,7 @@ class API:
     Attributes:
         logger (:obj:`Logger <teos.logger.Logger>`): The logger for this component.
         app: The Flask app of the API server.
+        stub (:obj:`TowerServicesStub`): The rpc client stub.
     """
 
     def __init__(self, inspector, internal_api_endpoint):
@@ -107,6 +108,8 @@ class API:
         self.app = Flask(__name__)
         self.inspector = inspector
         self.internal_api_endpoint = internal_api_endpoint
+        channel = grpc.insecure_channel(internal_api_endpoint)
+        self.stub = TowerServicesStub(channel)
 
         # Adds all the routes to the functions listed above.
         routes = {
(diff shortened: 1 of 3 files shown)


teos/api.py
```diff
@@ -99,6 +99,8 @@ class API:
     Attributes:
         logger (:obj:`Logger <teos.logger.Logger>`): The logger for this component.
         app: The Flask app of the API server.
+        channel (:obj:`grpc.Channel`): The :obj:`Channel` object.
+        stub: The rpc client stub.
```
sr-gi, line 103: Add type in docstring

teos/api.py
```diff
@@ -107,6 +109,8 @@ def __init__(self, inspector, internal_api_endpoint):
         self.app = Flask(__name__)
         self.inspector = inspector
         self.internal_api_endpoint = internal_api_endpoint
+        self.channel = grpc.insecure_channel(internal_api_endpoint)
```
sr-gi, line 112: Do we need to store the channel?

teos/rpc.py
```diff
@@ -67,6 +67,10 @@ class _RPC(TowerServicesServicer):
     Args:
         internal_api_endpoint (:obj:`str`): the endpoint where to reach the internal (gRPC) api.
         logger (:obj:`Logger <teos.logger.Logger>`): the logger for this component.
+
+    Attributes:
+        channel (:obj:`grpc.Channel`): the Channel object
+        stub (:obj:`TowerServicesStub`): the rpc client stub
```
sr-gi, line 73: Add type to docs

sr-gi (changes requested):
Minor comments.

In case of removing `self.channel`, that may apply to `rpc.py`, `api.py` and `teos_cli.py`.

teos/api.py
bigspider: Removed (also in rpc and teos_cli) in 5988290.

teos/api.py
bigspider: Added in 5988290.

sr-gi (approved):

PR #222, opened 2020-09-14
Modifies circle-ci to run docker builds only on tags

sr-gi:
Currently circle-ci is building docker images for all the distros for every commit on master. This is due to how the circle-ci filters work (like an or statement instead of an and).

Removes the check of branch being master and leaves the tag conditions.

For reference: https://stackoverflow.com/questions/58792796/circleci-ignores-git-tag-filters


bigspider (approved):
Looks good to me.

PR #225, opened 2020-09-14
Distinguishes between appointment and extended_appointment

sr-gi:
I've tried to do a better job distinguishing between `Appointment` and `ExtendedAppointment`.

I haven't changed the `Watcher`'s code much since I didn't want to change `self.appointments` to `self.extended_appointments`.

All appointments within the tower are extended anyway, whereas they are not on the user-side.

### DIFF
--- a/teos/builder.py
+++ b/teos/builder.py
@@ -30,14 +30,14 @@ class Builder:
         locator_uuid_map = {}
 
         for uuid, data in appointments_data.items():
-            appointment = ExtendedAppointment.from_dict(data)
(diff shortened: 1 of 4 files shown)


sr-gi:
@bigspider I'm happy to add any further change on the tower-side, but I was trying to avoid changing `self.appointments` for clarity.

bigspider:
LGTM, I think it's pretty clear as it's clearly documented in the watcher that it's an ExtendedAppointment, and there's no risk of mixing the two.

PR #228, opened 2020-09-15
Avoid hardcoded strings

bigspider:
Adds an AppointmentStatus Enum for the possible appointment statuses, and a class constant for the "END" message of the ChainMonitor.

A larger refactoring in the same spirit could make sense for the config dictionary keys, but it is not trivial because of the need of making it work with `config_loader.py`.

Also fixing a few documentation errors and typos

Closes: 173

### DIFF
--- a/common/appointment.py
+++ b/common/appointment.py
@@ -1,4 +1,12 @@
-import struct
+from enum import Enum
+
+
+class AppointmentStatus(str, Enum):
+    """Represents all the possible states of an appointment in the system, or in a response to a client request."""
+
+    NOT_FOUND = "not_found"
+    BEING_WATCHED = "being_watched"
+    DISPUTE_RESPONDED = "dispute_responded"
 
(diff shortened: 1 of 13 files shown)


common/appointment.py
```diff
@@ -1,4 +1,13 @@
 import struct
```
sr-gi, line 1: struct is not being used

sr-gi (changes requested):
Needs rebase and removing a couple of unused imports (there are some other minor style issues to be fixed, but I'd leave those for a final PR before tagging the first RC).

 `test/common/unit/test_appointment.py` is importing `struct` but not using it.

common/appointment.py
bigspider: Removed from both here and `test/common/unit/test_appointment.py`.

sr-gi (approved):

PR #229, opened 2020-09-15
Adds waitress as WSGI for Windows

sr-gi:
This PR removes Flask a a webserver option, and replaces it with `waitress`.

The original idea was to run `waitress` with `subprocess` so we can use the same approach used for `gunicorn`. I haven't been able to make it work with the command line runner (opened [an issue ](https://github.com/Pylons/waitress/issues/317) in the `waitress` repo about it).

`waitress` is used if the code is run on Windows. Currently I've enforced this by checking the OS version and modifying the config if it is misconfigured. This is basically because the defaults are different for Windows and Linux/OSX and didn't want to have multiple config params for them.

Another option would be to exit if the values are misconfigured but I think that would worsen the UX (specially for Windows, given `gunicorn` is the default).

In order to make the user more aware of issues (warnings and above), this PR also includes a slight modification of the logs to all non-info lines include the severity. Info lines are the most predominant so I haven't include it for simplicity.

Finally, given that flask is not an option anymore, `e2e` test are run using `gunicorn` if not on Windows.


teos/teosd.py
```diff
@@ -437,10 +441,10 @@ def main(config):
             if opt in ["--datadir"]:
                 data_dir = os.path.expanduser(arg)
             if opt in ["--wsgi"]:
-                if arg in ["gunicorn", "flask"]:
+                if arg in ["gunicorn", "waitress"]:
                     command_line_conf["WSGI"] = arg
                 else:
-                    exit("wsgi must be either gunicorn or flask")
```
bigspider, line 443: Commenting here in lack of a better place: L19 in /teos/help.py and the comment in L293 need to be updated to talk about `waitress` instead of `flask`.

bigspider (approved):
Only a note about a couple updated comments / help messages.

I tried to run it on Linux with "waitress" and it works fine for m; didn't try on Windows.

sr-gi:
Spent the morning trying to make this work on Windows, couldn't properly install leveldb. I've had this issue in the past and it was quite frustrating tbh.

Dependencies installed good nonetheless apart from that one, so given the minimal change I'm going to leave it as is and maybe give it another try once either we have Windows users or someone with more experience in Windows deb joins as a contributor.

PR #231, opened 2020-09-22
Handles request rejection for expired subscriptions

sr-gi:
Expired subscriptions were not being checked when data was being sent by / requested from the tower. This PR handled that case and also moves the `UserInfo` deletion to the `Gatekeeper` instead of it being handled by either the `Watcher` or the `Responder`.

For this changes to work, the `Gatekeeper` now needs to have a notion of time, which is given by making him subscribe to the `ChainMonitor`. Also, data is being deleted with a `USERS_CACHE_SIZE_BLOCKS` delay so both the `Watcher` and the `Responder` can query any user related data they may need in order to do their own cleanups.

### DIFF
--- a/teos/gatekeeper.py
+++ b/teos/gatekeeper.py
@@ -1,5 +1,13 @@
 from math import ceil
+from queue import Queue
 from threading import Lock
+from threading import Thread
+from readerwriterlock import rwlock
+from collections import OrderedDict
+
+from teos.cleaner import Cleaner
+from teos.chain_monitor import ChainMonitor
+from teos.constants import OUTDATED_USERS_CACHE_SIZE_BLOCKS
 
 from common.tools import is_compressed_pk, is_u4int
 from common.cryptographer import Cryptographer
@@ -11,8 +19,6 @@ from common.exceptions import InvalidParameter, InvalidKey, SignatureError
 class NotEnoughSlots(ValueError):
     """Raised when trying to subtract more slots than a user has available."""
 
-    pass
-
 
 class AuthenticationFailure(Exception):
     """
@@ -20,7 +26,9 @@ class AuthenticationFailure(Exception):
     not found within the registered ones.
     """
 
-    pass
+
+class SubscriptionExpired(ValueError):
+    """Raised when trying to subtract more slots than a user has available."""
 
 
 class UserInfo:
@@ -82,20 +90,48 @@ class Gatekeeper:
         user_db (:obj:`UsersDBM <teos.user_dbm.UsersDBM>`): A user database manager instance to interact with the
             database.
         registered_users (:obj:`dict`): A map of ``user_pk:user_info``.
+        outdated_users_cache (:obj:`dict`): A cache of outdated user ids to allow the Watcher and Responder to query
+            deleted data. Keys are bock heights, values are lists of user ids. Has a maximum size of
+            ``OUTDATED_USERS_CACHE_SIZE_BLOCKS``.
         lock (:obj:`Lock`): A lock object to lock access to the Gatekeeper on updates.
-
     """
 
     def __init__(self, user_db, block_processor, subscription_slots, subscription_duration, expiry_delta):
         self.subscription_slots = subscription_slots
         self.subscription_duration = subscription_duration
         self.expiry_delta = expiry_delta
+        self.block_queue = Queue()
         self.block_processor = block_processor
         self.user_db = user_db
         self.registered_users = {
             user_id: UserInfo.from_dict(user_data) for user_id, user_data in user_db.load_all_users().items()
         }
+        self.outdated_users_cache = {}
         self.lock = Lock()
+
+        # Starts a child thread to take care of expiring subscriptions
+        Thread(target=self.manage_subscription_expiry, daemon=True).start()
+
+    def manage_subscription_expiry(self):
+        """
+        Manages the subscription expiry of the registered users. Subscriptions are not deleted straightaway for two
+        purposes:
+
+        - First, it gives time to the ``Watcher`` and the ``Responder`` to query the necessary data for housekeeping,
+        and gives some reorg protection.
+        - Second, it gives a grace time to the user to renew their subscription before it is irrevocably deleted.
+        """
+
+        while True:
+            block_hash = self.block_queue.get()
+            # When the ChainMonitor is stopped, a final ChainMonitor.END_MESSAGE message is sent
+            if block_hash == ChainMonitor.END_MESSAGE:
+                break
+
+            # Expired user deletion is delayed. Users are deleted when their subscription is outdated, not expired.
+            block_height = self.block_processor.get_block(block_hash).get("height")
+            self.update_outdated_users_cache(block_height)
+            Cleaner.delete_outdated_users(self.get_outdated_user_ids(block_height), self.registered_users, self.user_db)
 
     def add_update_user(self, user_id):
         """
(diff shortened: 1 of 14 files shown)


teos/gatekeeper.py
```diff
@@ -83,20 +89,46 @@ class Gatekeeper:
...
             user_id: UserInfo.from_dict(user_data) for user_id, user_data in user_db.load_all_users().items()
         }
         self.lock = Lock()
 
+        # Starts a child thread to take care of expiring subscriptions
+        Thread(target=self.manage_subscription_expiry, daemon=True).start()
+
+    def manage_subscription_expiry(self):
+        """
+        Manages the subscription expiry of the registered users. When a subscription reaches it end, it moves it to
```
bigspider, line 112: Typo, "its end"

teos/gatekeeper.py
```diff
@@ -83,20 +89,46 @@ class Gatekeeper:
...
+        the expiry cache and deleted it after ``GATEKEEPER_USERS_CACHE`` blocks, giving time to the ``Watcher`` and
```
bigspider, line 113: typo "deletes it", I suppose.

teos/gatekeeper.py
```diff
@@ -83,20 +89,46 @@ class Gatekeeper:
...
+                break
+
+            with self.lock:
+                expired_users = self.get_expired_users(self.block_processor.get_block_count())
+                self.expired_users_cache.append(expired_users)
+                # Expired user deletion is delayed so the Watcher and Responder have enough time to query the associated
+                # data for the deletion of their data structures
+                if len(self.expired_users_cache) > USERS_CACHE_SIZE_BLOCKS:
+                    expired_users = self.expired_users_cache.pop(0)
+                    Cleaner.delete_expired_users(expired_users, self.registered_users, self.user_db)
```
bigspider, line 130: Is this still safe in case of a block re-org? I'd think it is, as long USERS_CACHE_SIZE_BLOCKS is significantly bigger than what we consider the re-org threshold (e.g. 10 >> 6). Agreed?

teos/watcher.py
```diff
@@ -294,10 +295,15 @@ def get_appointment(self, locator, user_signature):
...
             :obj:`AppointmentNotFound`: if the appointment is not found in the tower.
+            :obj:`SubscriptionExpired`: If the user subscription has expired.
         """
 
         message = "get appointment {}".format(locator).encode("utf-8")
         user_id = self.gatekeeper.authenticate_user(message, user_signature)
+        if self.gatekeeper.has_subscription_expired(user_id):
+            raise SubscriptionExpired(
+                f"Your subscription expired at {self.gatekeeper.registered_users[user_id].subscription_expiry}"
+            )
```
bigspider, line 306: Maybe better `"Your subscription expired at block {...}"`

teos/watcher.py
```diff
@@ -347,6 +354,10 @@ def add_appointment(self, appointment, user_signature):
             raise AppointmentLimitReached(message)
 
         user_id = self.gatekeeper.authenticate_user(appointment.serialize(), user_signature)
+        if self.gatekeeper.has_subscription_expired(user_id):
+            raise SubscriptionExpired(
+                f"Your subscription expired at {self.gatekeeper.registered_users[user_id].subscription_expiry}"
+            )
```
bigspider, line 360: Same as above, I'd mention "block"

test/teos/unit/test_api.py
```diff
@@ -189,8 +189,10 @@ def test_register_json_no_inner_dict(client):
 
 
 def test_add_appointment(internal_api, client, appointment, block_processor):
-    # Simulate the user registration (end time does not matter here)
-    internal_api.watcher.gatekeeper.registered_users[user_id] = UserInfo(available_slots=1, subscription_expiry=0)
+    # Simulate the user registration (end time does not matter here  as long as is in the future)
```
bigspider, line 192: extra space, and typo "as long as it is"

test/teos/unit/test_gatekeeper.py
```diff
@@ -196,6 +239,56 @@ def test_add_update_appointment(gatekeeper, generate_dummy_appointment):
         gatekeeper.add_update_appointment(user_id, appointment_uuid, appointment_x2_size)
 
 
+def test_has_subscription_expired(gatekeeper):
+    # Subscription are tagged as expired if the expiry + expiry_delta is grater than the current height
+    user_info = UserInfo(available_slots=1, subscription_expiry=gatekeeper.block_processor.get_block_count())
+    user_id = get_random_value_hex(32)
+    gatekeeper.registered_users[user_id] = user_info
+
+    # Check tha the subscription is still live
```
bigspider, line 248: typo "that"

test/teos/unit/test_gatekeeper.py
```diff
@@ -196,6 +239,56 @@ def test_add_update_appointment(gatekeeper, generate_dummy_appointment):
...
+
+
+def test_has_subscription_expired_not_registered(gatekeeper):
+    # If the users is unknown by the Gatekeeper, the method will fail
+    with pytest.raises(AuthenticationFailure):
+        gatekeeper.has_subscription_expired(get_random_value_hex(32))
+
+
+def test_get_expired_users(gatekeeper):
+    # Gets a list of users which subscription expires at a given height
```
bigspider, line 267: typo "whose subscription"

test/teos/unit/test_watcher.py
```diff
@@ -345,13 +347,32 @@ def test_add_appointment_no_slots(watcher, generate_dummy_appointment):
         watcher.add_appointment(appointment, appointment_signature)
 
 
+# FIXME: 194 will do with dummy appointment
+def test_add_appointment_expired_subscription(watcher, generate_dummy_appointment):
+    # Appointments from register users with expired subscriptions fail as well
```
bigspider, line 352: typo "registered users"

bigspider (approved):
Only a bunch of typos.

teos/gatekeeper.py
sr-gi: It is not actually, since the users in the cache are removed from `registered_users` without checking if the subscription is still expired after `USERS_CACHE_SIZE_BLOCKS`. Since the cache is not being reworked in a case of a reorg, it could be the case that it enters on a weird state. 

If we need to check `register_users` at `expiry + expiry_delta` and also `USERS_CACHE_SIZE_BLOCKS` blocks later, it may be way simpler to just drop items from the `register_users` at `expiry + expiry_delta + USERS_CACHE_SIZE_BLOCKS` and get rid of the cache.

That would mean that, effectively, we would be giving the user an additional `USERS_CACHE_SIZE_BLOCKS` windows where it may be able to renew the subscription and keep the data undeleted. I think that may be worth the simplicity of the approach tho.

sr-gi: More on this. Since we are already giving `expiry_delta` additional blocks to the user, we could rework the way that works. Subscription could expire exactly at `expiry`, and the repurpose `expire_delta` as the time we keep information around before deleting. That means that the user won't be able to query no send any data from the tower after `expiry`, but their data will be kept around `expiry_delta` more blocks, and then it will be permanently deleted.

That gets rid both of the cache and the additional constant.

teos/gatekeeper.py
sr-gi: Changes the description here, so it does not apply anymore.

teos/gatekeeper.py
sr-gi: Same as previous.

teos/watcher.py
sr-gi: Fixed.

teos/watcher.py
```diff
@@ -338,7 +344,8 @@ def add_appointment(self, appointment, user_signature):
         Raises:
             :obj:`AppointmentLimitReached`: If the tower cannot hold more appointments (cap reached).
             :obj:`AuthenticationFailure`: If the user cannot be authenticated.
-            :obj:`NotEnoughSlots`: If the user does not have enough available slots, so the appointment is rejected.
+            :obj:`NotEnoughSlots`: If the user does not have enough available slots, so the appointment is rejected.\
```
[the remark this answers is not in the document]
sr-gi: Fixed.

teos/watcher.py
sr-gi: Fixed.

test/teos/unit/test_api.py
sr-gi: Fixed.

test/teos/unit/test_gatekeeper.py
sr-gi: Fixed.

test/teos/unit/test_gatekeeper.py
sr-gi: Fixed.

test/teos/unit/test_watcher.py
sr-gi: Fixed.

sr-gi:
Reworked the subscription expiry to get rid of the cache as mentioned in https://github.com/talaia-labs/python-teos/pull/231#discussion_r492946714. I've added two separate commits that can be squashed once they're reviewed.

sr-gi:
Still missing tests for ``update_outdated_users_cache``, will add them tomorrow morning.

Needs rebasing to add all functionality in a single commit and all tests in another.

sr-gi:
Should be good now. Will need rebasing to add all functionality in a single commit and all tests in another once reviewed.

teos/gatekeeper.py
```diff
@@ -189,46 +228,117 @@ def add_update_appointment(self, user_id, uuid, ext_appointment):
...
-        expired_appointments = []
-        # Avoiding dictionary changed size during iteration
-        for user_id in list(self.registered_users.keys()):
-            if block_height == self.registered_users[user_id].subscription_expiry + self.expiry_delta:
-                expired_appointments.extend(self.registered_users[user_id].appointments)
+        outdated_users = self.get_outdated_users(block_height)
 
-        return expired_appointments
+        if block_height not in self.outdated_users_cache:
+            with self.lock:
```
bigspider, line 340: I would put all the code inside the lock, otherwise it could in principle happen that the code is stopped before getting the lock, meanwhile a reorg happens and a different block is processed, and the execution continues, while the state has changed. This would probably not be a problem here because the height after the re-org is likely the same, therefore the list of outdated users shouldn't change, but I think it's better to put the lock on the whole code and not have to think about it.

bigspider (changes requested):
Minor comments.

teos/gatekeeper.py
sr-gi: Added

PR #237, opened 2020-09-29
Lets the OS pick a free port instead of harcoding one

### LINKED ISSUE #236: Log server port cannot be hardcoded
Hardcoding the log server port means that more than one instance of the tower cannot run in the same machine (for instance one for mainnet and one for testnet).

We should get a distinct port for every instance.

sr-gi:
The current hardcoded approach prevents a user from running multiple tower on the same host (for different networks for instance). With this approach the limitation is fixed.

Close #236

### DIFF
--- a/teos/api.py
+++ b/teos/api.py
@@ -59,7 +59,7 @@ def get_request_data_json(request):
         raise InvalidParameter("Request is not json encoded")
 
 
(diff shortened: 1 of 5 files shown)


bigspider (approved):
Nice one.

Issue #203, opened 2020-08-21
Refactor cli with a library to reduce boilerplate


bigspider:
Adding a new command to the cli requires quite some boilerplate code, and it's easy to miss something. Using a library among the ones listed here might help to make code leaner and more robust: https://docs.python-guide.org/scenarios/cli/

PR #244, opened 2020-10-09
Refactor of the cli

bigspider:
Refactored the cli in order to make sure that each command (including its help message) is entirely implemented within a single class.

The base of all the cli commands is the `CliCommand` class. Each implementation must have a `name`, and defines the command and parameters of the command by specifying the values passed to `getopt`, which is the default behavior of the `parse_args` method; alternatively, the `parse_args` method could in principle be overridden to define an arbitrary other way of parsing the parameters.

The `Cli` class is the actual runner of all the commands. It provides a decorator that is applied to each command class to register all the commands. The decorator also does some basic sanity checks on the correctness of the subclass.

For each command (subclass of `CliCommand`), the docstring is exactly the text that is shown by the `help` command for it.
Moreover, the `show_help` function now generates the list of commands by parsing such docstring for each command. For this to work, the line containing "NAME:" in the docstring must be present and properly formatted. If that fails for any command, `show_help` will raise an error; a test was added to make sure that `show_help` does not raise.

Closes: #203, #245


teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+        command.
+    """
+
+    name = "help"
+
+    @staticmethod
+    def run(rpc_client, opts_args):
+        opts, args = opts_args
+
+        print(opts_args)
```
sr-gi, line 387: Remove this print?

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+    # where command is the first argument not starting with "-".
+
+    try:
+        cmd_index = next(i for i in range(len(argv)) if i > 0 and not argv[i].startswith("-"))
+    except StopIteration:
+        sys.exit(show_usage())
+
+    command = argv[cmd_index]
+    global_options = argv[1:cmd_index]
+    command_args = argv[cmd_index + 1 :]
```
sr-gi, line 412: I don't think this covers all the cases.

What if there is a command that has multiple arguments and also command options?

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+        It any error that might happen, showing an appropriate message to console.
+
+        Returns:
+            The return value of the ``run`` command, or :obj:`None` if an error was raised.
+        """
+
+        if command_name not in self.COMMANDS:
+            sys.exit("Unknown command. Use help to check the list of available commands")
+
+        cmd = self.COMMANDS[command_name]
```
sr-gi, line 249: May be worth calling it command? cmd reminds me to the command line interface (may apply to other places).

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+
+        try:
+            args = cmd.parse_args(raw_args)
+            return self.COMMANDS[command_name].run(self.rpc_client, args)
```
sr-gi, line 253: You have assigned `self.COMMANDS[command_name]` to `cmd`, why not use it?

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+        except grpc.RpcError as e:
```
sr-gi, line 254: **Check global comment first**

If we create a custom exception we'll be able to remove grpc from here.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
-    rpc_client = RPCClient(teos_rpc_host, teos_rpc_port)
+    @classmethod
+    def parse_args(cls, args):
+        """Parses the ``args`` array using ``getopt``, using ``shortopts`` and ``longopts`` as options."""
 
-    result = None
-    try:
-        if command == "get_all_appointments":
-            result = rpc_client.get_all_appointments()
+        return getopt(args, cls.shortopts, cls.longopts)
```
sr-gi, line 175: Am I missing something here, or you're not using the opts in any way?

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
 
-    setup_data_folder(data_dir)
+    All the implementations should have an appropriately formatted docstring. See existing commands for an example.
+    Any implementation _must_ override the ``name`` attribute, and it might override the ``shortopts`` and ``longopts``
+    attributes.
+    """
 
-    teos_rpc_host = config.get("RPC_BIND")
-    teos_rpc_port = config.get("RPC_PORT")
+    name = None
```
sr-gi, line 167: I think this variables could be part of `__init__`, and parse_args won't be a class method.

sr-gi (changes requested):
First pass, I like the approach even if it's a pretty custom way of building a CLI.

Some general notes:

- Given we are making this way more modular, the `RPCClient` could be moved to it's own file now, alongside `to_json` and `formatted`. That would move all the RPC imports and functionality to there. (We may still need to import `grpc` to handle the `grpc.RpcError` in `Cli`, but that can either be left as is, or create a custom exception and wrap the raise around that, so we can import it from the `teos_cli.py`.
- If `run` is going to be removed, `setup.py` needs to be updated so the console script calls `main` instead.
- This will need proper testing now that it can be.

teos/cli/teos_cli.py
bigspider: Yes, this class is basically meant to be like an abstract class (initially I tried with the `abc` module, but then removed it as all the decorators were just adding tons of additional boilerplate, with little apparent benefit) .
`CLICommand` calls the `parse_args` method and passes the result as an argument to the corresponding `run` method (of the subclass).

teos/cli/teos_cli.py
bigspider: Uhm, I think that would change the design. My idea here is that the subclasses of `CLICommand` are never instantiated, so there are only static or class methods. `parse_args` in `CLICommand` gives a reasonable default behavior (parsing with `getopt`, but subclasses might in principle override that with a completely different `parse_args`, which should nevertheless be stateless. This is also necessary if we want to register the commands with a decorator.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
         print("Closing the Eye of Satoshi")
 
 
-def main(command, args, data_dir, command_line_conf):
-    # Loads config and sets up the data folder and log file
-    config_loader = ConfigLoader(data_dir, CONF_FILE_NAME, DEFAULT_CONF, command_line_conf)
-    config = config_loader.build_config()
+class CliCommand:
+    """
+    Base class of each Cli command.
```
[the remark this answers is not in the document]
bigspider: Changed to CLI everywhere in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
         print("Closing the Eye of Satoshi")
 
 
-def main(command, args, data_dir, command_line_conf):
-    # Loads config and sets up the data folder and log file
-    config_loader = ConfigLoader(data_dir, CONF_FILE_NAME, DEFAULT_CONF, command_line_conf)
-    config = config_loader.build_config()
+class CliCommand:
```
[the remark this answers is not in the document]
bigspider: Done in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+        raise NotImplementedError()
 
-        elif command == "get_user":
-            if not args:
-                sys.exit("No user_id was given")
-            if len(args) > 1:
-                sys.exit(f"Expected only one argument, not {len(args)}")
 
-            result = rpc_client.get_user(args[0])
+class Cli:
```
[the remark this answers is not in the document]
bigspider: Done in 26dcfcf591ed3c63b0aae95a524007493b8044ce

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+        teos_rpc_port = config.get("RPC_PORT")
 
-                elif command == "stop":
-                    sys.exit(help_stop())
+        self.rpc_client = RPCClient(teos_rpc_host, teos_rpc_port)
 
-                else:
-                    sys.exit("Unknown command. Use help to check the list of available commands")
+    @classmethod
+    def _register_command(cls, command_cls):
```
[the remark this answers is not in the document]
bigspider: changed to simply `command` in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
-    result = None
-    try:
-        if command == "get_all_appointments":
-            result = rpc_client.get_all_appointments()
+        return getopt(args, cls.shortopts, cls.longopts)
 
-        elif command == "get_tower_info":
-            result = rpc_client.get_tower_info()
+    @classmethod
+    def run(rpc_client, opts_args):
```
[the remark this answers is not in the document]
bigspider: Indeed, I changed this few times and didn't realize it was wrong in the base class. 26dcfcf591ed3c63b0aae95a524007493b8044ce
I changed it to `@staticmethod` to be equal to all the subclasses; not sure I have a good reason to choose between `static` and `class` method, so I used the most limiting by default, can be changed later if the need arises.

teos/cli/teos_cli.py
```diff
@@ -110,85 +155,267 @@ def stop(self):
...
+
+    DESCRIPTION:
+
+        Gets information about all appointments stored in the tower.
+    """
+
+    name = "get_all_appointments"
+
+    @staticmethod
+    def run(rpc_client, opts_args):
```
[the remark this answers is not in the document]
bigspider: As mentioned above, instead I changed it in the base class to be a `@staticmethod` as well. Now the signature matches, so I guess that covers it.

teos/cli/teos_cli.py
```diff
@@ -209,18 +436,22 @@ def run():
...
-            sys.exit("No command provided. Use help to check the list of available commands")
-        else:
-            sys.exit("Unknown command. Use help to check the list of available commands")
-
     except GetoptError as e:
         sys.exit("{}".format(e))
 
+    if command in Cli.COMMANDS:
+        try:
+            cli = Cli(DATA_DIR, command_line_conf)
```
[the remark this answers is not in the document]
bigspider: Fixed in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
bigspider: Oops, removed in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
bigspider: Done in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
bigspider: Done in 26dcfcf591ed3c63b0aae95a524007493b8044ce.

teos/cli/teos_cli.py
bigspider: I moved RPCClient to a separate file, but didn't create a custom exception yet; I don't quite see a large benefit in doing so, and it would mean wrapping every call from the RPCClient to handle RpcError and reraise the custom exception.

teos/cli/teos_cli.py
bigspider: Seems good to me. Do you have a failure example?
My assumption here is that the format is:
```
  teos-cli [global_options] command [command options and/or arguments]
```
where all the global_options must start with 1 or two `-`, and `command` definitely does not.
Since `cmd_index` is the first command line arg not starting with `-`, and the parsing of everything after `command` is delegated to the `CLICommand` itself, this seems sound to me.

bigspider:
Should have addressed most of the requested changes.
Renamed `main` to `run` so the entrypoint is consistent with the teos and the client, rather than fixing `setup.py`.

I added some tests for the CLI class; basically they are just mocking the rpc_client and checking that it is correctly proxying requests there. If the behavior of the CLI becomes more complex, it might make sense to remove all the `sys.exit()` calls from the CLI, and instead use return values (or errors) to detect various conditions; for now, it didn't seem worth it.

There are no tests for the code inside `run()` yet; it could make sense to also refactor the argument parsing inside a method in order to make it testable. Let me know if you think it should be done before merging this one, or I'll open a new ticket.

teos/cli/teos_cli.py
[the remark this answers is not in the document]
sr-gi: AFAIK static uses the class in some way, without changing nor depending on any instantiation of it, whereas class is for constructors.

teos/cli/teos_cli.py
sr-gi: Oh OK, so whatever comes after the command is passed to the specific command for custom parsing, right?

teos/cli/teos_cli.py
sr-gi: Then you can at least remove `shortopts` and `longopts`, since they are never used. The latter is also setup to default for `getopt`, so doing:

```
return getopt(args, "")
```

Should do.

test/teos/unit/cli/test_teos_cli.py
```diff
@@ -0,0 +1,92 @@
+import os
+import pytest
+import sys
```
sr-gi, line 3: `sys` is never used.

test/teos/unit/cli/test_teos_cli.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_show_usage_does_not_throw():
+    # If any of the Cli commands' docstring has a wrong format and cannot be parsed, this will raise an error
+    show_usage()
+
+
+def test_cli_init_does_not_throw():
+    try:
+        cli = CLI(".teos-cli-test", {})
```
sr-gi, line 23: No need to store it

test/teos/unit/cli/test_teos_cli.py
```diff
@@ -0,0 +1,92 @@
...
+def test_get_users(cli, monkeypatch):
+    rpc_client_mock = MagicMock(cli.rpc_client)
+    monkeypatch.setattr(cli, "rpc_client", rpc_client_mock)
+
+    cli.run("get_users", [])
+
+    rpc_client_mock.get_users.assert_called_once()
+
+
+def test_stop(cli, monkeypatch):
```
sr-gi, line 86: test_stop is defined twice

sr-gi (changes requested):
Looks better now. Some additional comments:

- There are still places using `cmd` or `cmd_*`.
- I think more testing is required:
    - The RPCClient is not tested for example, nor are some of the negative or command specific cases (e.g. test get_user with an invalid userid). 
    - `CLI` in general is almost non-tested for invalid return types. 
    - ...
Covering the run method and not exiting straightaway can be covered on a followup I think.

teos/cli/teos_cli.py
sr-gi: The are still some matches for `cmd*`

teos/cli/teos_cli.py
bigspider: Yeah, that's the idea, the Command's `parse_args` will take care of the rest.

sr-gi: OK, will need to do a bit of testing but looks good

teos/cli/teos_cli.py
bigspider: All the methods of CLICommand and subclasses must be either `static` or `class` method, as no instances are ever created, so there's no value bound to `self`.
`@abstractmethod` would require to use the `abc` package (especially for the CLICommand's `run` method), which I had initially tried, but didn't see to bring any advantage at the end; which is perhaps not surprising as there are no instances aver created, so not much point in calling things "abstract".

I can see why `@classmethod` is typically used for factory methods, as it would give access to the constructors without breaking in case of inheritance (which is impossible if you use a `@staticmethod`); but accessing the class variables from subclasses (`shortopts` and `longopts` here) seems a valid use case as well.

test/teos/unit/cli/test_teos_cli.py
bigspider: Removed assignment in 5267444f9044725c10fb19b89962b33f22d454e9.

test/teos/unit/cli/test_teos_cli.py
bigspider: Oops, I was trying to use it in an early attempt for a test. Removed it in 5267444f9044725c10fb19b89962b33f22d454e9.

test/teos/unit/cli/test_teos_cli.py
bigspider: Removed in aad0a93b6815ba0b58ccdd1328a22aeb3897cc95.

teos/cli/teos_cli.py
```diff
@@ -1,194 +1,328 @@
...
-        """
-        Gets information about a specific user.
+        self.rpc_client = RPCClient(teos_rpc_host, teos_rpc_port)
 
-        Args:
-            user_id (:obj:`str`): the id of the requested user.
+    @classmethod
+    def command(cls, command_cls):
+        """
+        Register a new command, which must be a subclass of :class:`CLICommand` and must override the ``name`` field
```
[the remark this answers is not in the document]
bigspider: Added in aad0a93b6815ba0b58ccdd1328a22aeb3897cc95.

bigspider:
Opened a separate issue #245 for the RPCClient. Not super urgent imho, as most of the commands just proxy to the actual rpc, so not a lot to test there. Might be more interesting to have fully end-to-end tests where some commands are done via the cli.

I suppose some more testing could be done for `CLI.run` (especially failure cases), but all the tests for commands are actually calling CLI.run that dispatches to the command; therefore the positive case is tested. To test the failure cases, I suppose we should refactor it so that the `sys.exit` is not called from within `CLI`.

sr-gi:
I think #245 could be tackled here. May even be worth also adding the sys.exit bit so we can add proper testing to the whole thing.

This would need squashing btw. I would suggest current code can go in one, given it is pretty self contained, and test in another one. The sys.exit bit could go in a separate one since it builds on top of the code improvements.

teos/cli/teos_cli.py
```diff
@@ -209,18 +343,22 @@ def run():
...
         sys.exit("{}".format(e))
 
+    if command in CLI.COMMANDS:
+        try:
+            cli = CLI(data_dir, command_line_conf)
+            result = cli.run(command, command_args)
+            if result:
+                print(result)
+        except InvalidParameter:
+            sys.exit(show_usage())
```
sr-gi, line 356: This should first print the exception message, and then `sys.exit` Since we are casting everything to BasicException (and InvalidParameter inherits from it), here we could simply do:

```
print(e.msg if not e.kwargs else f"{e.msg}. Error arguments: {e.kwargs}")
sys.exit(show_usage())
```

teos/cli/teos_cli.py
```diff
@@ -1,194 +1,328 @@
...
-    setup_data_folder(data_dir)
+        try:
+            args = command.parse_args(raw_args)
+            return command.run(self.rpc_client, args)
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.UNAVAILABLE:
+                return "It was not possible to reach the Eye of Satoshi. Are you sure the tower is running?"
+            else:
+                return e.details()
+        except InvalidParameter as e:
```
sr-gi, line 168: I'd remove this `InvalidParameter` catch and let it be captured by the outer scope.

teos/cli/teos_cli.py
```diff
@@ -1,194 +1,328 @@
...
+            return e.msg if not e.kwargs else f"{e.msg}. Error arguments: {e.kwargs}"
+        except Exception as e:
+            return f"Unknown error occurred: {str(e)}"
```
sr-gi, line 171: Here we could create a `BasicException` so everything can be captured with `except BasicException` in the outer scope.

test/teos/e2e/test_cli_e2e.py
```diff
@@ -0,0 +1,117 @@
...
+    # Get back its info
+    user = json.loads(rpc_client.get_user(user_id))
+
+    assert set(user.keys()) == set(["appointments", "available_slots", "subscription_expiry"])
+
+
+def test_get_user_non_existing(teosd, rpc_client):
+    # Get a user that does not exist
+    with pytest.raises(InvalidParameter):
+        json.loads(rpc_client.get_user("00112233445566778899aabbccddeeff00112233445566778899aabbccddeeff00"))
```
sr-gi, line 117: What's the point of json.loads here if `rpc_client.get_user` will fail?

test/teos/e2e/test_cli_e2e.py
```diff
@@ -0,0 +1,117 @@
...
+    _, teos_id = teosd
+
+    # Check that there is no appointment, so far
+    all_appointments = json.loads(rpc_client.get_all_appointments())
+    watching = all_appointments.get("watcher_appointments")
+    responding = all_appointments.get("responder_trackers")
+    assert len(watching) == 0 and len(responding) == 0
+
+    # Register a user
+    available_slots, subscription_expiry = teos_client.register(user_id, teos_id, teos_base_endpoint)
```
sr-gi, line 50: The return values are never used

test/teos/e2e/test_cli_e2e.py
```diff
@@ -0,0 +1,117 @@
...
+    assert set(tower_info.keys()) == set(
+        ["n_registered_users", "tower_id", "n_watcher_appointments", "n_responder_trackers"]
+    )
+
+
+def test_get_users(teosd, rpc_client):
+    _, teos_id = teosd
+
+    # Register a user
+    available_slots, subscription_expiry = teos_client.register(user_id, teos_id, teos_base_endpoint)
```
sr-gi, line 95: The return values are never used

test/teos/e2e/test_cli_e2e.py
```diff
@@ -0,0 +1,117 @@
...
+    users = json.loads(rpc_client.get_users())
+
+    assert user_id in users
+
+
+def test_get_user(teosd, rpc_client):
+    _, teos_id = teosd
+
+    # Register a user
+    available_slots, subscription_expiry = teos_client.register(user_id, teos_id, teos_base_endpoint)
```
sr-gi, line 106: Here you could tests that the values on register match the ones returned, not only that the keys of the returned dict match

test/teos/e2e/test_cli_e2e.py
sr-gi: PS: Same as bellow, check that the data also matches, not just that the id is on the returned data

test/teos/unit/cli/test_rpc_client.py
```diff
@@ -0,0 +1,17 @@
+import pytest
```
sr-gi, line 1: I know I told you to test the rpc_client, but is there any point on testing it like this? Just wondering.

test/teos/e2e/conftest.py
```diff
@@ -35,6 +35,9 @@ def teosd():
     teosd_process.join()
     shutil.rmtree(".teos")
 
+    # FIXME: wait some time, otherwise it might fail when multiple e2e tests are ran in the same session. Not sure why.
+    sleep(1)
```
sr-gi, line 39: Can you share the error you get? I've tried to reproduce but I cannot.

Not that it's a big deal, but if we can find the source of the issue it may be nicer.

sr-gi (changes requested):
Some git housekeeping, small tweaks and some nits.

Should be mergeable after this.

- There are several files here being change for rearranging comments / docs. This should go to a separate PR so we don't mess with history in unrelated commits. Feel free to open an issue to keep track of what we see that is not properly formatted and we can go and do it all in bulk. For this specific PR, I've marked the ones for this PR so they can be rebased out.

- Modify lines 61 and 114 for `teos_cli.py` to remove the references to logging, given there is no logging in the cli anymore. Add them to 855ad21d588dc4591c0abca14c402173f09d9c2a.

- Regarding tests:
    - The last two commits (without taking the linting into account) are a bit intertwined. I think they can either be squased or split so one covers unit and the other covers e2e. If they are separated,the one covering unit may be squashable with e3a3250b7d7b8525abc0c7bf47c722a1f0196087.
    - In `cli_e2e`, you could have a global variable keeping track of what has been sent to the tower, so you could later check that the data queried to the tower is correct (in `test_get_tower_info` for example). Since this is being kept track by global variables, it should work no matter if the tests are run independently or as a suite.
    - Also in `cli_e2e`, since you are registering users to check they are in the tower response once queried, why not check they are not there before the registration (for completeness basically).

test/teos/unit/cli/test_rpc_client.py
bigspider: Well, the `get_user` function has some logic beyond just proxying to the rpc service (raising InvalidParameter), so I think it makes sense to have a unit test for it, I think.

test/teos/e2e/test_cli_e2e.py
bigspider: Indeed, removed in 37ef745fe6a74b3cdad0a95c96b029c4ec59e57f.

test/teos/e2e/test_cli_e2e.py
bigspider: Good point, added in 37ef745fe6a74b3cdad0a95c96b029c4ec59e57f.

test/teos/e2e/test_cli_e2e.py
bigspider: `get_users` only returns a list of user ids, so not sure how to do it without also using `get_user`.
In 37ef745fe6a74b3cdad0a95c96b029c4ec59e57f I removed the return values instead.

test/teos/e2e/test_cli_e2e.py
bigspider: Removed unused variables in 37ef745fe6a74b3cdad0a95c96b029c4ec59e57f.

test/teos/e2e/conftest.py
bigspider: I get failures like these if I run all the tests with `pytest test/teos`:

```
=========================================================================================
FAILED test/teos/unit/test_api.py::test_register - AssertionError: assert '029235274c47...9e142fff932f9' == '0365239fcb5a...ea98163f74f6c'
FAILED test/teos/unit/test_api.py::test_register_top_up - AssertionError: assert '029235274c47...9e142fff932f9' == '0365239fcb5a...ea98163f74f6c'
FAILED test/teos/unit/test_api.py::test_add_appointment - AssertionError: assert 99 == 0
FAILED test/teos/unit/test_api.py::test_add_appointment_registered_no_free_slots - assert 200 == 400
```

I'd guess some unit tests are running before the e2e tests had a chance to complete cleanup (in fact, `test_api` is the first unit test module in lexicographical order).

test/teos/unit/cli/test_teos_cli.py
```diff
@@ -101,15 +101,6 @@ def test_run_exception(cli, monkeypatch):
     assert "Unknown error occurred: Mock Exception" == cli.run("mock_command_exception", [])
 
 
-def test_get_tower_info(cli, monkeypatch):
```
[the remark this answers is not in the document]
bigspider: The test in a14a2a3 is an e2e test; here it's removed because there were two copies of the same test.

teos/cli/teos_cli.py
bigspider: Actually, I'm not sure this fits the current design. CLI.run's documentation says that `If any error is raised by the command, returns an error message.`, so it wouldn't make sense to have an exception for InvalidParameter, imho.

If anything, the error handling in in the main `run()` method is redundant if we keep this design.
Rather, I think the CLI.run() method should return the concatenation of the error message and the result of show_usage(), if we want to show the usage for InvalidParameter.

teos/cli/teos_cli.py
bigspider: Same as per comment above about handling in the outer scope.

teos/cli/teos_cli.py
bigspider: Done in 9e4b087, although this is pending some other changes as per comment below.

teos/cli/teos_cli.py
```diff
@@ -1,194 +1,328 @@
...
+    name = "get_user"
+
+    @staticmethod
+    def run(rpc_client, opts_args):
+        opts, args = opts_args
+
+        if not args:
+            return "No user_id was given"
+        if len(args) > 1:
+            return f"Expected only one argument, not {len(args)}"
```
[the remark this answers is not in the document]
bigspider: Done in 9e4b087.

teos/cli/teos_cli.py
```diff
@@ -1,194 +1,328 @@
...
+    """
+
+    name = "get_user"
+
+    @staticmethod
+    def run(rpc_client, opts_args):
+        opts, args = opts_args
+
+        if not args:
+            return "No user_id was given"
```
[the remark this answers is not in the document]
bigspider: Done in 9e4b087.

test/teos/unit/cli/test_teos_cli.py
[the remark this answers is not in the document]
sr-gi: OK, fair enough.

teos/cli/teos_cli.py
sr-gi: That's fine, if you rather sanitise in `.run` and return the user feedback to `main` I'm OK with it.

teos/cli/teos_cli.py
sr-gi: OK

teos/cli/teos_cli.py
sr-gi: Replied bellow, this may need rebase then.

test/teos/e2e/test_cli_e2e.py
sr-gi: Oh, my bad, mixed it up with `get_user`.

test/teos/unit/cli/test_rpc_client.py
sr-gi: Oh OK, fair enough.

test/teos/e2e/conftest.py
sr-gi: Ohh, gotcha. I was only trying it out running the e2e suite, not the whole test suite. Let me see if I can find what's wrong.

teos/cli/teos_cli.py
bigspider: Done in 6b8b5975d4c2884280767a97117c7f181ce2041e (which is rebased to merge with the previous commit).

bigspider:
I was getting lost with trying to get a meaningful history, so I instead squashed all the work on tests into 6f03da7d453f7206d6a36dc09832d7d33e572139, including the previous changes from PR review.

About the comments in the review:

> In cli_e2e, you could have a global variable keeping track of what has been sent to the tower, so you could later check that the data queried to the tower is correct (in test_get_tower_info for example). Since this is being kept track by global variables, it should work no matter if the tests are run independently or as a suite.

The numbers for `n_watcher_appointments` and `n_responder_trackers` can easily be tracked only if at the end of each test you make sure that all the appointments go back to zero, as they change when appointments are triggered or expire; hard to keep track reliably imho and it makes the code of tests hard to track. We could track n_registered_users, but currently I'm just registering one user anyway (and to make sure the correct answer is "1", we would have to register the user within `test_get_tower_info` as well (or rely on other tests calling it before, while now each test in test_cli_e2e can be successfully ran individually.
Tbh, I'm also generally uncomfortable with using global variables in tests, unless they are basically constants for the whole module.
If there is any changing state, it should be encapsulated somewhere, and not managed from within tests, as that makes tests very hard to follow.


> Also in cli_e2e, since you are registering users to check they are in the tower response once queried, why not check they are not there before the registration (for completeness basically).

The registration is done with teos client, not with the cli; so I think this test would be out of place in cli_e2e.

test/teos/e2e/conftest.py
sr-gi: Cannot reproduce, but not like it is a big deal anyway, so I guess we can leave it like this for now.

teos/appointments_dbm.py
```diff
@@ -29,7 +29,7 @@ class AppointmentsDBM(DBManager):
     Args:
         db_path (:obj:`str`): the path (relative or absolute) to the system folder containing the database. A fresh
             database will be created if the specified path does not contain one.
-            
+
```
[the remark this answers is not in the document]
sr-gi: Opened an issue to keep track of these: https://github.com/talaia-labs/python-teos/issues/255

teos/chain_monitor.py
```diff
@@ -86,8 +86,8 @@ def __init__(self, receiving_queues, block_processor, bitcoind_feed_params):
 
     def enqueue(self, block_hash):
         """
-        Adds a new block hash to the internal queue of the  :obj:`ChainMonitor` and the internal state. The state contains
-        the list of ``last_tips`` to prevent notifying about old blocks. ``last_tips`` is bounded to
+        Adds a new block hash to the internal queue of the  :obj:`ChainMonitor` and the internal state. The state
+        contains the list of ``last_tips`` to prevent notifying about old blocks. ``last_tips`` is bounded to
```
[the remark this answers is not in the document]
sr-gi: Opened an issue to keep track of these: https://github.com/talaia-labs/python-teos/issues/255

teos/watcher.py
```diff
@@ -673,8 +673,8 @@ def get_user_info(self, user_id):
             user_id (:obj:`str`): the id of the requested user.
 
         Returns:
-            :obj:`UserInfo <teos.gatekeeper.UserInfo> or :obj:`None`: The user data if found. :obj:`None` if not found, or
-            the ``user_id`` is invalid.
+            :obj:`UserInfo <teos.gatekeeper.UserInfo> or :obj:`None`: The user data if found. :obj:`None` if not found,
```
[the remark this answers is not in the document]
sr-gi: Opened an issue to keep track of these: https://github.com/talaia-labs/python-teos/issues/255

sr-gi (approved):
LGTM

PR #249, opened 2020-10-14
Reworks teos to use RWLocks

sr-gi:
This basically add RWLocks to:

- `Gatekeeper`
- `Responder`
- `Watcher`

Before this PR, the `InternalAPI` has a RWLock to guard the access to the backend. However, two threads are running for each of the main backend components (`Watcher` and `Responder`), meaning that, theoretically, their `do_watch` method could crash with the user data feeding (`add_appointment` and `handle_breach`). This tries to fix that.

For the `Gatekeeper`:

The writer lock is acquired for the data addition/update (`add_update_*` and `update_outdated_users_cache`) and deletion (`delete_appointments`). The reader lock is acquired for data querying (`authenticate_user`, `has_subscription_expired` and  `get_outdated_users`).

For the `Responder`:

The writer lock is acquired for `add_tracker` (called by `handle_breach`) and for a part of `do_watch` and `handle_reorgs`. Notice that the whole `do_watch` cannot be covered in a single piece since `handle_reorgs` may call `handle_breach` and lead to a deadlock.

For the `Watcher`:

The writer lock is acquired for the most part of `add_appointment` and `do_watch`, whereas the reader lock is acquired for get requests.

The `Watcher`'s block cache is maintaining its own RWLock, even though it could arguably be removed.

### DIFF
--- a/teos/responder.py
+++ b/teos/responder.py
@@ -131,6 +131,7 @@ class Responder:
         block_processor (:obj:`BlockProcessor <teos.block_processor.BlockProcessor>`): A block processor instance to
             get data from bitcoind.
         last_known_block (:obj:`str`): The last block known by the :obj:`Responder`.
+        rw_lock (:obj:`RWLockWrite <rwlock.RWLockWrite>`): A lock object to manage access to the Responder on updates.
     """
 
     def __init__(self, db_manager, gatekeeper, carrier, block_processor):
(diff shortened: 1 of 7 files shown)


sr-gi:
@bigspider this may be a bit tricky to review

teos/watcher.py
```diff
@@ -228,6 +227,7 @@ def __init__(self, db_manager, gatekeeper, block_processor, responder, sk, max_a
         self.signing_key = sk
         self.last_known_block = db_manager.load_last_block_hash_watcher()
         self.locator_cache = LocatorCache(blocks_in_cache)
+        self.rw_lock = rwlock.RWLockWrite()
```
bigspider, line 230: Should add attribute in the class docs.
Same for `LocatorCache` (there it was already missing before this PR).

teos/responder.py
```diff
@@ -144,6 +145,7 @@ def __init__(self, db_manager, gatekeeper, carrier, block_processor):
         self.carrier = carrier
         self.block_processor = block_processor
         self.last_known_block = db_manager.load_last_block_hash_responder()
+        self.rw_lock = rwlock.RWLockWrite()
```
bigspider, line 148: Add attribute to class docs.

bigspider (commented):
It looks much better indeed. Only thing I think is still non optimal design-wise is that the watcher is explicitly accessing the lock of the responder and gatekeeper. Perhaps it would be cleaner (and less bug-prone) to have additional accessors for whatever the watcher is accessing, and those accessors would take care of the lock, so all the locking is taken explicit care within the same class defining the lock (meaning: basically treat rw_lock as a private member). What do you think?

sr-gi:
Yeah, I was trying to avoid adding getters for those, but I guess it actually makes sense from a cleaner design POV.

bigspider (approved):
Looks good to me.

PR #250, opened 2020-10-14
Integrate wsgi logs

sr-gi:
Integrates the gunicorn and waitress logs with the logging server.

The approach is pretty straightforward, but a bit tricky. gunicorn allows a python config file to rewrite how it handles logs. The provided config file `gunicorn_conf.py` adds the timestamp and component to the message in a similar way used by `teos.logger`. The tricky parts comes with `SockerHandler`, since it does not use formatters. 

In order to work around it, `FormattedSocketHandler` is created inheriting from `SockerHandler` and applying the desired formatting before emitting the message. 

For waitress it is a bit easier, since `setup_logging` can be called. However, since it does not use `structlog` some additional config is needed to add the missing parts, in a similar fashion to `gunicorn` (`FormattedSocketHandler` + Timestamper + `add_component`).

### DIFF
--- a/teos/api.py
+++ b/teos/api.py
@@ -85,7 +85,9 @@ def serve(internal_api_endpoint, endpoint, logging_port, min_to_self_delay, auto
     api.logger.info(f"Initialized. Serving at {endpoint}")
 
     if auto_run:
-        wsgi_serve(api.app, listen=endpoint)
+        # Waitress will serve both on IPV4 and IPV6 if localhost is passed as endpoint. Defaulting to IPV4 only in that
+        # case.
+        wsgi_serve(api.app, listen=endpoint.replace("localhost", "127.0.0.1"))
     else:
         return api.app
 
--- a/teos/logger.py
+++ b/teos/logger.py
@@ -1,6 +1,7 @@
 import logging
 import logging.config
 import logging.handlers
+from logging.handlers import SocketHandler
 import structlog
 import pickle
 import socketserver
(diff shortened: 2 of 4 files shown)


bigspider:
Tried to search a bit to find a solution to send gunicorn logs apropriately to the stream handler, and I don't have a clean one yet. This might have some useful info: https://albersdevelopment.net/2019/08/15/using-structlog-with-gunicorn/

A dirty solution could be to identify and parse separately the logs coming from gunicorn from within our `LogRecordStreamHandler.handle_log_record`, which could be done by checking the LogRecord's name (I'm getting `gunicorn.error`), or simply by switching to a different output when record.msg is not json and treat them separately, something like:

```python
    @staticmethod
    def handle_log_record(record):
        """
        Processes log records received via the socket. The record's ``msg`` field is expected to be an encoded
        :obj:`dict` produced by :obj:`StructLog`. The :obj:`dict` is encoded to a string using ``encode_event_dict`` and
        sent to the logger with the name specified in the record.

        Args:
            record (:obj:`logging.LogRecord`): a log record.
        """
        try:
            event_dict = json.loads(record.msg)
        except json.JSONDecodeError:
            # TODO: parse non-json record nicely
            event_dict = {"event": record.msg}

        event_dict.update({"log_level": record.levelname})
        message = encode_event_dict(event_dict)

        logger = logging.getLogger(record.name)
        logger.log(record.levelno, message, exc_info=record.exc_info, stack_info=record.stack_info)
```

sr-gi:
Took a look at the post you referred to, was pretty helpful but looks like there still something missing regarding SocketHandler.

The approach works for StreamHandler when outputting to console, but not when sending the data via the socket. In the latter only the event is sent instead of the whole struct message.

I'm guessing it is due to some kind of wrapper missing from the current configuration, but couldn't figure out what to add to the config dict to make it work, given that we do it in a different way for the logger.

sr-gi:
Added Waitress logs since it was easy enough.

It is not working in the current version of waitress, but a PR was recently merged (https://github.com/Pylons/waitress/pull/315) for the upcoming release that will make it work.

teos/logger.py
```diff
@@ -14,20 +15,52 @@
...
+        return record.name in self.names
+
+
+def add_component(logger, name, event_dict):
+    """
+    Adds the component to the entry if the entry is not from structlog. This only applies to the API since it is
+    handled by external WSGI servers.
+    """
+    event_dict["component"] = "API"
+    return event_dict
```
bigspider, line 63: I think it would be better to also specialized the name of the method to `add_api_component`, as the current name suggests a more generic method.

teos/logger.py
```diff
@@ -14,20 +15,52 @@
...
 
-class JsonMsgLogger(logging.Logger):
+
+class FormattedSocketHandler(SocketHandler):
+    """ Works in the same way as SocketHandler, but it uses formatters. """
+
+    def emit(self, record):
+        """
+        Works exactly like SocketHandler.emit but formats the record before sending it.
+        record.args is set to none since they are already used by self.format, otherwise makePickle would try to
```
bigspider, line 27: Nit in docs: `none` => `None`.

teos/api.py
```diff
@@ -84,7 +84,8 @@ def serve(internal_api_endpoint, endpoint, logging_port, min_to_self_delay, auto
     api.logger.info(f"Initialized. Serving at {endpoint}")
 
     if auto_run:
-        wsgi_serve(api.app, listen=endpoint)
+        # Waitress will serve both on IPV4 and IPV6 if localhost is passed as endpoint
```
bigspider, line 87: Comment is a bit unclear. Is the purpose of the `replace` to only serve on IPV4 instead?

bigspider (approved):
Having tried myself, I think this was a tough one in terms of putting all the pieces together!

Looks good to me, only noted some minutiae.

teos/api.py
sr-gi: Yeah. If localhost is passed as a param, waitress would serve on two endpoints: `127.0.0.1` and `::1`.

There's no need for it to serve on both, so I changed it to default to IPV4, since I don't think it may be an issue for a local server.

teos/logger.py
sr-gi: Changed to `add_api_component`

teos/api.py
sr-gi: Clarified comment in 40399e4

Issue #232, opened 2020-09-23
Add get_subscription_details command to client


sr-gi:
A useful RPC for the client would be `get_subscription_details`, to check the number of remaining slots and subscription expiry time, as well as to compare it with its local view.

orbitalturtle:
Going to give this a whirl

sr-gi:
Closed via #251

sr-gi:
Closed via #251

PR #251, opened 2020-10-18
Getsubscriptioninfo

orbitalturtle:
Adds a get_subscription_info rpc command to the client as described in #232


teos/api.py
```diff
@@ -314,3 +315,48 @@ def get_appointment(self):
...
+        except InvalidParameter as e:
+            self.logger.info("Received invalid get_subscription_info request", from_addr="{}".format(remote_addr))
+            return jsonify({"error": str(e), "error_code": errors.INVALID_REQUEST_FORMAT}), HTTP_BAD_REQUEST
+
+        self.logger.info("Received get_subscription_info request", from_addr="{}".format(remote_addr))
+
+        try:
+            r = self.stub.get_subscription_info(GetSubscriptionInfoRequest(signature=request_data.get("signature")))
+
+            response = {
```
sr-gi, line 348: Protobuf encodes numeric values as floats when using `Struct`, so we need to do a little tweak here since the returned numeric values are always int (and float would make no sense for expiry nor remaining slots). Notice this is only relevant because you are using `GetUserResponse` for the response of the RPC call, which uses `Struct`.


```
response = intify(
    json_format.MessageToDict(r.user, including_default_value_fields=True, preserving_proto_field_name=True)
)
```

You'll need to import `intify` from `common.tools`.

teos/api.py
```diff
@@ -314,3 +315,48 @@ def get_appointment(self):
...
+                    r.user, including_default_value_fields=True, preserving_proto_field_name=True
+                )
+            }
+            rcode = HTTP_OK
+
+        except grpc.RpcError as e:
+            rcode = HTTP_BAD_REQUEST
+            response = {
+                "error": e.details(),
+                "error_code": errors.APPOINTMENT_INVALID_SIGNATURE_OR_INSUFFICIENT_SLOTS,
```
sr-gi, line 359: This should be `APPOINTMENT_INVALID_SIGNATURE_OR_SUBSCRIPTION_ERROR`. It was renamed recently but left unchanged in `get_appointment`. #247 is correcting that.

contrib/client/teos_client.py
```diff
@@ -463,6 +499,18 @@ def main(command, args, command_line_conf):
...
+            if args:
+                arg_opt = args.pop(0)
+
+                if arg_opt in ["-h", "--help"]:
+                    sys.exit(help_get_subscription_info())
+
+            teos_id = load_teos_id(config.get("TEOS_PUBLIC_KEY"))
+            subscription_info = get_subscription_info(user_sk, teos_id, teos_url)
+            if subscription_info:
+                print(subscription_info)
```
sr-gi, line 512: It may be worth pretty printing this (this may also apply to `appointment_data`). This should also use logger.info instead of print.

Feel fee to leave this as a followup.

sr-gi: BTW by pretty printing I mean `json.dumps(x, indent=4)`.

We do so in the CLI but looks like we forgot to port it here too.

teos/internal_api.py
```diff
@@ -158,6 +158,35 @@ def get_appointment(self, request, context):
...
+            try:
+                subscription_info = self.watcher.get_subscription_info(request.signature)
+
+                user_struct = Struct()
+
+                user_struct.update(
+                    {
+                        "subscription_expiry": subscription_info.subscription_expiry,
+                        "available_slots": subscription_info.available_slots,
+                        "appointments": list(subscription_info.appointments.keys()),
```
sr-gi, line 173: Watch out here. `appointments.keys()` are the appointment `uuids`, not the locators. Those ids are internal and unknown by the user. You'll need to pull the locators from the subscription info here.

sr-gi (changes requested):
First pass through the code, I think it looks pretty good. I haven't looked at the tests yet, will do after the fixes. Most inline comments are nits / style.

Some general comments:

- `teos_client needs black formatting`.
- general help needs a new entry for `get_subscription_info`.
- `appointment_pb2, tower_services_pb2, user_pb2` need black formatting.
- Add E2E tests to check that the received data matches the sent one (check apppoitment uuid comment at `InternalAPI).

The thing I most concerned about this command is that a malicious user doing MITM could reuse the request message for authentication given it has no parameters, so learning it once it's enough. A secure and authenticated channel is a MUST anyway when talking to the tower, but I guess we would need to add more emphasis about it in the docs to make sure no-one runs without HTTPS.

orbitalturtle:
Hey @sr-gi, pushed up those updates. Just couldn't figure out how to black format the proto files. Ran black --line-length=120 on them but that didn't do anything

teos/api.py
```diff
@@ -314,3 +316,46 @@ def get_appointment(self):
...
+        """
+        Gives information about a user's subscription from the watchtower.
+
+        Only a user who has registered can retrieve this information. User must provide the correct signature.
+
+        Returns:
+            :obj:`str`: A json formatted dictionary containing information about the user's subscription.
+
+            Returns not found if the user is not yet registered with the watchtower.
+
```
sr-gi, line 330: Remove unnecessary line break at the end of the docs

teos/watcher.py
```diff
@@ -673,11 +673,44 @@ def get_user_info(self, user_id):
...
+        if self.gatekeeper.has_subscription_expired(user_id):
+            raise SubscriptionExpired(
+                f"Your subscription expired at block {self.gatekeeper.registered_users[user_id].subscription_expiry}"
+            )
+
+        subscription_info = self.gatekeeper.registered_users.get(user_id)
+
+        locators = []
+
+        for appt_uuid in subscription_info.appointments:
```
sr-gi, line 707: There's no need to go to the DB for this, you can check `self.appointments.get(uuid)` and get the locator (if it is on the `Watcher`). If the data is not there, it must be on the `Responder`, so `self.responder.trackers.get(uuid)`.

For both the field `locator` should be set.

sr-gi (changes requested):
Looks pretty good now. Basically nits + an improvement to access the data from the backend.

Do not worry about concurrent access to this since we're already taking care or it at #249, so I'll rebase it to also include this case once merged.

Regarding tests, it'll be nice to check that the expiry and slots also match our local view, so you could, for instance, `get_subscription_info`, fill some slots with appointments, and `get_subscription_info` again to see that it matches.

Also do unit test the watcher (there are currently no tests for `get_subscription_info`), to make sure there is no unhandled case that may make this crash. And if there is, that it cannot be triggered from a request.

orbitalturtle:
@sr-gi Thanks again for the review! Pushed up the latest changes.

teos/watcher.py
```diff
@@ -673,11 +673,47 @@ def get_user_info(self, user_id):
...
+            if appt_uuid in self.appointments:
```
sr-gi, line 708: Not exactly. You're returning all the data here, when only the locator was needed.

`self.appointments.get(appt_uuid).get("locator")` and `self.responder.trackers.get(appt_uuid).get("locator")` should do the trick.

Furthermore, I'd add an else with a log.debug recording that the locator for the uuid was not found. This should not happen, but if it does at least we'll have logs for it.

test/teos/e2e/test_basic_e2e.py
```diff
@@ -88,13 +99,22 @@ def test_commands_registered(run_bitcoind):
     assert r.get("locator") == appointment.locator
     assert r.get("appointment") == appointment.to_dict()
     appointments_in_watcher += 1
+    available_slots -= 1
+
+    # Get subscription info
+    r = get_subscription_info()
+    assert r.get("appointments")[0].get("locator") == appointment.locator
```
sr-gi, line 106: Once you fix the issue with the return, `.get("localtor")` won't be needed here.

test/teos/e2e/test_basic_e2e.py
```diff
@@ -187,6 +209,15 @@ def test_multiple_appointments_life_cycle(run_bitcoind):
         add_appointment(appointment)
         appointments_in_watcher += 1
 
+    available_slots -= 5
```
sr-gi, line 212: This part can get a bit hard to maintain if some of the params of the function are changed. 

A better approach would be to decrease the number of available slots one by one in the previous for loop.

test/teos/e2e/test_basic_e2e.py
```diff
@@ -187,6 +209,15 @@ def test_multiple_appointments_life_cycle(run_bitcoind):
...
+
+    # Check that get_subscription_info also detects these new appointments and returns the correct info.
+    r = get_subscription_info()
+    # We've added 6 appointments so far
+    assert len(r.get("appointments")) == 6
```
sr-gi, line 217: Same here, this gets hard to maintain given it is hardcoded, there are currently two variables keeping track of how many appointments are in the watcher and the responder, this should be equal to the sum of both.

test/teos/unit/test_watcher.py
```diff
@@ -757,3 +757,27 @@ def test_filter_breaches_random_data(watcher, generate_dummy_appointment):
...
+
+    message = "get subscription info"
+    signature = Cryptographer.sign(message.encode("utf-8"), user_sk)
+
+    sub_info = watcher.get_subscription_info(signature)
+
+    assert len(sub_info.appointments) == 0
+
+    # Now let's try sending an invalid signature with the wrong message signed.
+    message = "meh"
```
sr-gi, line 779: Feel like this last bit could be moved to `test_internal_api` as a `test_get_appointment_info_non_registered`, since that tests is missing. Here I think you can keep the previous part and the just add some additional data to `appointments` / `self.responder.trackers` and check that it matches.

sr-gi (changes requested):
Last couple of details to be fixed.

This will also need rebase btw, it's behind master.

test/teos/unit/test_api.py
```diff
@@ -554,3 +555,22 @@ def test_get_appointment_in_responder(internal_api, client, generate_dummy_track
     assert tx_tracker.dispute_txid == r.json.get("appointment").get("dispute_txid")
     assert tx_tracker.penalty_txid == r.json.get("appointment").get("penalty_txid")
     assert tx_tracker.penalty_rawtx == r.json.get("appointment").get("penalty_rawtx")
+
+
+def test_get_subscription_info(internal_api, client, generate_dummy_tracker, monkeypatch):
```
sr-gi, line 560: The `generate_dummy_tracker` fixture is not necessary here.

test/teos/unit/test_api.py
```diff
@@ -554,3 +555,22 @@ def test_get_appointment_in_responder(internal_api, client, generate_dummy_track
...
+    # Need to mock a user
+    internal_api.watcher.gatekeeper.registered_users[user_id].available_slots = 100
```
sr-gi, line 562: You'll need to register here since otherwise this test won't work standalone. Notice that the current approach won't work given that the user entry for `user_id` does not exist if the other tests are not run.

`internal_api.watcher.gatekeeper.add_update_user(user_id)` should do.

Then if you want to modify the registration data to something custom, feel free to do so.

test/teos/unit/test_api.py
```diff
@@ -554,3 +555,22 @@ def test_get_appointment_in_responder(internal_api, client, generate_dummy_track
...
+    monkeypatch.setattr(internal_api.watcher.gatekeeper, "authenticate_user", mock_authenticate_user)
+
+    # Request back the data
+    message = "get subscription info"
+    signature = Cryptographer.sign(message.encode("utf-8"), user_sk)
+    data = {"signature": signature}
+
+    # Next we can request the subscription info
+    r = client.post(get_subscription_info_endpoint, json=data)
+    assert r.status_code == HTTP_OK
```
sr-gi, line 576: It'll be nice to check that what you get here matches what you set on the registration.

sr-gi (changes requested):
Some final improvements on the tests I missed on my last review. Sorry about that.

test/teos/unit/test_internal_api.py
```diff
@@ -245,6 +252,16 @@ def test_get_appointment_non_registered(internal_api, stub, generate_dummy_appoi
     response = stub.get_appointment(GetAppointmentRequest(locator=appointment.locator, signature=request_signature))
     assert isinstance(response, GetAppointmentResponse)
 
+    # Now let's try sending an invalid signature with the correct user key, but the wrong message signed.
```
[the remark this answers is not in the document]
orbitalturtle: Ah no worries!

sr-gi:
LGTM. I've added some minor fixes on my fork https://github.com/sr-gi/python-teos/tree/getsubscriptioninfo (mainly `black` + some commit amending). Take a look at the diff, and if you're OK with those I can push them here.

orbitalturtle:
@sr-gi looks good ~

sr-gi (approved):

PR #254, opened 2020-11-03
plugin: Bumps cryptography requirement due to discovered vulnerability #253

sr-gi:
close #253

### DIFF
--- a/watchtower-plugin/requirements.txt
+++ b/watchtower-plugin/requirements.txt
@@ -1,7 +1,7 @@
 pyln-client
 requests
 coincurve
-cryptography==2.8
+cryptography>=2.8
 pyzbase32
 plyvel
 backoff


watchtower-plugin/requirements.txt
```diff
@@ -1,8 +1,8 @@
 pyln-client
 requests
 coincurve
-cryptography==2.8
+cryptography>=2.8
```
Legogris, line 4: I assume the vulnerability mentioned is the one addressed in `3.2`? If so, shouldn't this be `>=3.2` as the previous 2.8 still satifies this

sr-gi: Hey @Legogris, thanks for the ping about this.

That is indeed the vulnerability mentioned. I think I set it to `>=` so it gets the latest version, but didn't really bump the minimum since the vulnerability seems to be rather tangential to our use of the library. We only use `cryptography` for `ChaCha20Poly1305`.

However, this made me realize that the requirements are actually outdated, since the only module that currently needs the `cryptography` module is `common`. I'll double check that everything works as expected with 3.2 and update all requirement files.

PR #270, opened 2021-03-04
Fix e2e tests so they work with bitcoind 0.21.0

orbitalturtle:
When I tried running the e2e tests, they would hang indefinitely. While debugging I found that bitcoind would get caught at bitcoin_cli.getnewaddress() in /test/teos/conftest.py, getting the error: `A default wallet is no longer automatically created`

I'm running version 0.21.0 of bitcoin, and it looks like in version 0.21.0 of bitcoind "Bitcoin Core will no longer automatically create new wallets on startup": https://lists.linuxfoundation.org/pipermail/bitcoin-core-dev/2021-January/000097.html

Proposing the following change, which allows tests to run smoothly while running the latest version of bitcoind

### DIFF
--- a/test/teos/conftest.py
+++ b/test/teos/conftest.py
@@ -70,6 +70,11 @@ def setup_node():
             )
         )
         try:
+            info = bitcoin_cli.getnetworkinfo()
+
+            if info.get("version") >= 210000:
+                bitcoin_cli.createwallet("test_wallet")
+
             btc_addr = bitcoin_cli.getnewaddress()
             break
 


test/teos/conftest.py
```diff
@@ -70,7 +70,12 @@ def setup_node():
             )
         )
         try:
-            btc_addr = bitcoin_cli.getnewaddress()
+            info = bitcoin_cli.getnetworkinfo()
+
+            if info.get("version") >= 210000:
+                bitcoin_cli.createwallet("test_wallet")
+            
+            btc_addr = bitcoin_cli.getnewaddress("test_wallet")
```
sr-gi, line 78: I think the label here is optional, can you double check?

sr-gi (approved):
Agreed, I've noticed this when testing #267.

test/teos/conftest.py
orbitalturtle: Ah yup runs fine without that. Pushed up that change

sr-gi:
Hey @orbitalturtle, looks like this commit was not properly signed so I cannot merge the PR.

Can you sign and push again?

(Given you have yo rebase, feel free to rebase the branch to 9852fe95d087cdff3d13beb5c3adb191daa882b0 and remove my merge commit 9b408a0).

orbitalturtle:
@sr-gi Cool fixed that & the formatting issue ~ I forgot to set up signed commits on my new desktop!

PR #271, opened 2021-03-09
teos - Adds bitcoind polling to ensure it is alive

sr-gi:
The ChainMonitor polling thread now is also used to make sure bitcoind has not crashed. Locks have been added so the ChainMonitor thread can lock other background threads if bitcoind is down. This does not apply to user requests, since they need to give a timely response to the user.

In a nutshell, every time a thread makes a (blocking) call to bitcoind, if a `ConnectionRefusedError` is returned, the lock will be cleared and the function will be called recursively (this second time being blocked by the lock).

On the other hand, the `ChainMonitor` polling thread is the one in charge of setting the lock once the connection with bitcoind is reestablished, unblocking all other thread if applicable.

There are calls that should always be blocking, like the ones made by the `Carrier` or the ones makes by the `do_watch` threads of the `Watcher` and the `Responder` (trough the `BlockProcessor`). Therefore, the methods of the latter have a flag to indicate whether the query should be blocking or not. An example of non-blocking queries is a request from the user, that should be hold until the connection is reestablish but fail and notify the user that the service is unavailable, so it can try again later.

It's worth mentioning that bitcoind crashes should be rather uncommon, but this was still a non-handled edge case.

### DIFF
--- a/teos/api.py
+++ b/teos/api.py
@@ -169,8 +169,12 @@ class API:
                 response["public_key"] = user_id
 
             except grpc.RpcError as e:
-                rcode = HTTP_BAD_REQUEST
-                response = {"error": e.details(), "error_code": errors.REGISTRATION_MISSING_FIELD}
+                if e.code() == grpc.StatusCode.UNAVAILABLE:
+                    rcode = HTTP_SERVICE_UNAVAILABLE
+                    response = {"error": e.details()}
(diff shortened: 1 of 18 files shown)


sr-gi:
@orbitalturtle would you mind taking a look to double check it makes sense?

Issue #246, opened 2020-10-13
Teos-client register  timeout


orbitalturtle:
Hi there. I'm running into an issue when I try to register an appointment from the teos-client.

It seems that the API receives the register request, but then the request times out.

Here are the logs from the server:

13/10/2020 18:26:20 [API] Received register request  (from_addr=127.0.0.1)
[2020-10-13 14:26:50 -0400] [31893] [CRITICAL] WORKER TIMEOUT (pid:31902)
[2020-10-13 14:26:51 -0400] [31902] [INFO] Worker exiting (pid: 31902)
[2020-10-13 14:26:51 -0400] [31930] [INFO] Booting worker with pid: 31930
13/10/2020 18:26:51 [API] Initialized. Serving at localhost:9814

sr-gi:
Umm, never seen that one before. Looks like `gunicorn` is timing out indeed.

Can you add debug log level to https://github.com/talaia-labs/python-teos/blob/master/teos/teosd.py#L269-L277 so hopefully we can have a more detailed trace? Like so:

```
self.api_proc = subprocess.Popen(
    [
        "gunicorn",
        f"--bind={api_endpoint}",
        f"teos.api:serve(internal_api_endpoint='{self.internal_api_endpoint}', "
        f"endpoint='{api_endpoint}', logging_port='{logging_port}', "
        f"min_to_self_delay='{self.config.get('MIN_TO_SELF_DELAY')}')",
        "--log-level debug",
    ]
)
```

Then run the tower, try to register and add the logs from the point you start:

```
13/10/2020 20:16:53 [Daemon] Tower identity found. Loading keys
[...]
```

orbitalturtle:
sure thing, unfortunately doesn't look like it adds any more info:

> 14/10/2020 05:33:31 [Daemon] Tower identity found. Loading keys
> 14/10/2020 05:33:31 [Daemon] tower_id = 0230b267961623a4d82264c5aee3fa3fc77fe093cdcaa9b411aed94dc92b2fb8ed
> 14/10/2020 05:33:31 [Daemon] Starting TEOS
> 14/10/2020 05:33:31 [Daemon] Fresh bootstrap
> 14/10/2020 05:33:31 [Daemon] Internal API initialized. Serving at localhost:50051
> [2020-10-14 01:33:31 -0400] [11998] [INFO] Starting gunicorn 20.0.4
[...]

PR #272, opened 2021-03-09
Patches #246

sr-gi:
Temporary fix for #246.

defaults GRPC_DNS_RESOLVER to native until a fix for https://github.com/grpc/grpc/issues/24018 is found.

### DIFF
--- a/teos/teosd.py
+++ b/teos/teosd.py
@@ -403,6 +403,9 @@ def run():
     # between process.
     multiprocessing.set_start_method("spawn")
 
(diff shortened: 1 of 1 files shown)


sr-gi:
@orbitalturtle @3nprob could you give it a try?

If you can tell me what OS / version were you running the tower in so I can try to reproduce with and without the patch it may also help.

orbitalturtle:
I *think* I was running Ubuntu 18.04, will see if I can double check

3nprob:
Yup, this solves it for me

PR #297, opened 2021-07-08
Fix .teos volume path in Dockerfile

3nprob:
As of containerd 1.4.6, the relative path in the volume will cause the container to not run

https://github.com/containerd/containerd/issues/5547

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,12 +1,12 @@
 FROM python:3
-VOLUME ["~/.teos"]
+VOLUME ["/root/.teos"]
 WORKDIR /srv
 ADD . /srv/python-teos
 RUN apt-get update && \
     apt-get -y --no-install-recommends install libffi-dev libssl-dev pkg-config libleveldb-dev && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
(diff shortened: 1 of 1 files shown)


sr-gi:
Feels like it's a problem with containerd not with out sintax, doesn't it? Anyway, we're assuming this is run as root in the container, so I guess it's not a big deal to change it to a global path.

Two minor things:

- It'll be nice if you can rebase instead of merging master into the branch to avoid unnecessary commits into master.
- We do require signed commits. Would you mind signing it? https://github.com/talaia-labs/python-teos/blob/master/CONTRIBUTING.md#signing-commits

3nprob:
@sr-gi Sorry for leaving this hanging! I did rebase and sign the commit with a GPG key added to my github profile. Unfortunately it seems GH shows "unverified" because I don't put a verified e-mail address as author in the commit message (despite e-mail matching in key and commit).

I did put my GPG fingerprint in my GH bio, FWIW.

Is this all right for you or do I need to figure out some new e-mail for this purpose? I like to avoid having e-mail publicly easily discoverable for various reasons and it's not ideal if someone legit tries to contact me over an e-mail address I'd never check.

sr-gi:
That's good, thanks for taking the time for doing this.

sr-gi (approved):

